PR #225, opened 2021-01-07
feat(StakingInput): show insufficient sUSD on burn to target

evgenyboxer:


translations/en.json
```diff
@@ -228,8 +228,7 @@
 					"empty": "Enter an amount",
 					"burn": "Burn <0>sUSD</0>",
 					"max": "Max",
-					"to-target": "600%",
-					"insufficient": "Insufficient sUSD",
```
evgenyboxer, line 232: those two are not used ^

sections/staking/components/StakingInput/StakingInput.tsx
```diff
@@ -134,10 +137,24 @@ const StakingInput: React.FC<StakingInputProps> = ({
 					{isMint ? t('staking.actions.mint.action.empty') : t('staking.actions.burn.action.empty')}
 				</StyledCTA>
 			);
+		} else if (
+			burnType === BurnActionType.TARGET &&
+			maxBurnAmount != null &&
+			burnAmountToFixCRatio != null &&
+			burnAmountToFixCRatio.isGreaterThan(maxBurnAmount)
```
evgenyboxer, line 144: this is the main condition that needs to be checked ^

sections/staking/components/BurnTiles/BurnTiles.tsx
```diff
@@ -17,6 +25,8 @@ type BurnTilesProps = {
 };
 
 const burnIcon = <Svg src={BurnCircle} />;
+const burnCustomIcon = <Img src={BurnCustomCircle} />; // TODO: investigate why it doesn't render correctly with <Svg /> (ids were replaced to be unique)
```
andytcf, line 28: @evgenyboxer could it be the same issue as you told me, the mask-type

andytcf (approved):
LGTM

sections/staking/components/BurnTiles/BurnTiles.tsx
evgenyboxer: doesn't seem so... after more testing, it seems that it conflicts with `burn-circle.svg` but I couldn't find any matching ids...

PR #313, opened 2021-02-08
Feat/debt overview

clementbalestrat:


hooks/useHistoricalDebtData.ts
```diff
@@ -0,0 +1,85 @@
...
+					issuanceDebt: historicalIssuanceAggregation[i],
+					actualDebt: debtSnapshot.debtBalanceOf,
+				});
+			});
+
+			// Last occurrence is the current state of the debt
+			// Issuance debt = last occurrence of the historicalDebtAndIssuance array
+			historicalDebtAndIssuance.push({
+				timestamp: new Date().getTime(),
+				actualDebt: toBigNumber(debtDataQuery.data?.debtBalance ?? 0).toNumber(),
```
andytcf, line 73: don't convert to number, store as bigNumber and when displaying either use .toString() or convert to number then

hooks/useHistoricalDebtData.ts
```diff
@@ -0,0 +1,85 @@
...
+				issuanceDebt: last(historicalIssuanceAggregation) ?? 0,
+			});
+
+			setHistoricalDebt(historicalDebtAndIssuance);
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
andytcf, line 79: which deps are missing here, should prob include them and not to use eslint-disable anymore

pages/debt.tsx
```diff
@@ -0,0 +1,109 @@
...
+const TotalSynthValue = styled(StatBox)`
+	.title {
+		color: ${(props) => props.theme.colors.purple};
+	}
+`;
+
+const Header = styled.div`
+	color: ${(props) => props.theme.colors.white};
+	font-family: ${(props) => props.theme.fonts.extended};
+	font-size: 16px;
```
andytcf, line 100: change font-size to 14px, i don't think we have 16 anywhere on the dApp

sections/debt/DebtChart/DebtChart.tsx
```diff
@@ -0,0 +1,159 @@
...
+	);
+};
+
+type Data = {
+	issuanceDebt: number;
+	actualDebt: number;
+};
+
+const DebtChart = ({ data }: { data: Data[] }) => {
+	const { colors, fonts } = useContext(ThemeContext);
```
andytcf, line 65: ```const theme = useTheme``` would suffice as its apart of styled-components documentation


```import { useTheme } from 'styled-components';```

File: sections/debt/DebtChart/DebtChart.tsx [hunk 5]
```diff
@@ -0,0 +1,159 @@
...
+	if (!data || data.length === 0) return null;
```
andytcf, line 66: can we use a spinner for the edge case

```if(data.loading)```

andytcf (changes requested):
Left some comments

pages/debt.tsx
clementbalestrat: Wrong. 16px is used for infoBlock heading

sections/debt/DebtChart/DebtChart.tsx
clementbalestrat: Good call

PR #317, opened 2021-02-10
/synth page decimals fix

rockefeller92:
noticed that some USD values displayed on the /synths page had too many decimals and it was confusing me. dug a little deeper and realized it stemmed from the Currency.Amount component using the same currencyKey to determine how many decimals to use for both crypto and fiat amounts. updated Currency.Amount to accept an amountCurrencyKey and valueCurrencyKey and updated AssetsTable to provide the valueCurrencyKey separately

before:
Capture

after:
Capture2


clementbalestrat:
Thanks @rockefeller92. Good find!

PR #324, opened 2021-02-12
get the correct balancer lp token price for APR and TVL calcs

andytcf:


queries/liquidityPools/helper.ts
```diff
@@ -23,3 +25,24 @@ export async function getCurveTokenPrice(): Promise<number> {
...
+				where: {
+					id: `\\"${sTSLAPoolTokenAddress}\\"`,
+				},
+			},
+			properties: ['id', 'liquidity', 'totalShares'],
+		},
+		max: 1,
+		// @ts-ignore
+	}).then((result) => {
+		return Number(result[0].liquidity) / Number(result[0].totalShares);
```
clementbalestrat, line 46: What if there is no results? Do you have a `catch` or something to handle that?

clementbalestrat (approved):
LGTM, just a minor comment.

PR #331, opened 2021-02-15
refactor(Timer): Switch to a more human readable format

### LINKED ISSUE #319: Make epoch timer "e.g 00:00:02:33" more clear and less confusing by specifying the labels "hhs, mins" etc
The epoch time is too confusing and it is difficult to tell what the numbers represent, minutes get confused with hours etc

evgenyboxer:
Fixes #319

### DIFF
--- a/constants/date.ts
+++ b/constants/date.ts
@@ -1,3 +1,5 @@
 import getISOWeeksInYear from 'date-fns/getISOWeeksInYear';
 
 export const WEEKS_IN_YEAR = getISOWeeksInYear(new Date());
+
+export const DURATION_SEPARATOR = ' ';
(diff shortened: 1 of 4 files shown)


constants/date.ts
```diff
@@ -1,3 +1,5 @@
 import getISOWeeksInYear from 'date-fns/getISOWeeksInYear';
 
 export const WEEKS_IN_YEAR = getISOWeeksInYear(new Date());
+
+export const DURATION_SEPARATOR = ' ';
```
evgenyboxer, line 5: I think this is more readable than having the `:`

clementbalestrat (approved):

PR #353, opened 2021-02-20
sub-menu

clementbalestrat:


sections/shared/Layout/SideNav/SideNav.tsx
```diff
@@ -20,17 +21,25 @@ import { CryptoCurrency, Synths } from 'constants/currency';
 import { SIDE_NAV_WIDTH, zIndex } from 'constants/ui';
 
 import { MENU_LINKS, MIGRATE_MENU_LINKS } from '../constants';
+import SubMenu from './SubMenu';
 import PriceItem from './PriceItem';
 import PeriodBarStats from './PeriodBarStats';
 import BalanceItem from './BalanceItem';
 import CRatioBarStats from './CRatioBarStats';
 
+const getKeyValue = <T extends object, U extends keyof T>(obj: T) => (key: U) => obj[key];
```
clementbalestrat, line 30: To fix a TS error with objects and indices.

sections/shared/Layout/SideNav/SideNav.tsx
```diff
@@ -20,17 +21,25 @@ import { CryptoCurrency, Synths } from 'constants/currency';
...
+
 const SideNav: FC = () => {
 	const { t } = useTranslation();
 	const { asPath } = useRouter();
+	const menuLinkItemRefs = useRef({});
```
clementbalestrat, line 35: Using refs here to catch the `y` position of every menu item, so we can then pass it to a props and use `top: y px` so menu and sub menu can be aligned

sections/shared/Layout/SideNav/SideNav.tsx
```diff
@@ -20,17 +21,25 @@ import { CryptoCurrency, Synths } from 'constants/currency';
...
 	const SNX24hrPricesQuery = useSNX24hrPricesQuery();
 	const cryptoBalances = useCryptoBalances();
 	const synthsBalancesQuery = useSynthsBalancesQuery();
+	const [subMenuConfiguration, setSubMenuConfiguration] = useState({
```
clementbalestrat, line 39: As explained above, a subMenu configuration is passed to `SubMenu` with two infos: 
- and array with all the routes
- the `y` position it should be displayed

sections/shared/Layout/SideNav/SideNav.tsx
```diff
@@ -51,22 +60,46 @@ const SideNav: FC = () => {
 	const menuLinks = totalBalancePendingMigration > 0 ? MIGRATE_MENU_LINKS : MENU_LINKS;
 
 	return (
-		<SideNavContainer>
+		<SideNavContainer
```
clementbalestrat, line 63: if the cursor leaves the container then we hide the SubMenu

sections/shared/Layout/SideNav/SideNav.tsx
```diff
@@ -51,22 +60,46 @@ const SideNav: FC = () => {
...
+					<Svg src={StakingLogo} />
 				</Link>
 			</StakingLogoWrap>
 			<MenuLinks>
-				{menuLinks.map(({ i18nLabel, link }) => (
+				{menuLinks.map(({ i18nLabel, link, subMenu }, i) => (
 					<MenuLinkItem
+						ref={(r) => {
+							if (subMenu) {
+								menuLinkItemRefs.current = { ...menuLinkItemRefs.current, [i]: r };
```
clementbalestrat, line 76: For every `MenuItem` which contain `subMenu` we save their ref to `menuLinkItemRefs`

sections/shared/Layout/SideNav/SideNav.tsx
```diff
@@ -51,22 +60,46 @@ const SideNav: FC = () => {
...
+							}
+						}}
+						onMouseEnter={() => {
+							setSubMenuConfiguration(
+								subMenu
+									? {
+											routes: subMenu as any,
```
clementbalestrat, line 83: When a MenuItem is hovered we get its `boundingClientRect().y` and save it to the `subMenuConfiguration` object

sections/shared/Layout/SideNav/SubMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+type SubMenuProps = {
+	config: SubMenuConfiguration;
+	currentPath: string;
+};
+
+const SubMenu: FC<SubMenuProps> = ({ config, currentPath }) => {
+	const { t } = useTranslation();
+	return typeof document !== 'undefined'
+		? createPortal(
```
clementbalestrat, line 25: Using Portal here as the SideNav has a `fixed` position, which means there is no way to put the subMenu below with `z-index`. The only way is to send this component to the root.

sections/shared/Layout/SideNav/SubMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+				<Container isVisible={!!config?.routes ?? false}>
+					<Inner>
+						<FloatingContent topPosition={config?.topPosition ?? 0}>
+							<LinkList>
+								{config?.routes?.map(({ i18nLabel, subLink }) => (
+									<MenuLinkItem isActive={currentPath === subLink}>
```
evgenyboxer, line 31: Missing `key` right?

clementbalestrat: yep good find

sections/shared/Layout/SideNav/SubMenu.tsx
evgenyboxer: mice!

sections/shared/Layout/SideNav/SideNav.tsx
evgenyboxer: I wonder if it's because `menuLinkItemRefs` and `subMenuConfiguration` need to be typed

evgenyboxer (approved):
@clementbalestrat great stuff!

I think typing the ref/state might help with the TS errors, can be fixed later.

sections/shared/Layout/SideNav/SideNav.tsx
clementbalestrat: Yeah I think so too.

PR #358, opened 2021-02-24
Show renBTC/wBTC and ETH/wETC as part of debt overview

mi-yu:
@clementbalestrat I've added renBTC/wBTC to the sBTC totals and wETH/ETH to sETH in the debt portfolio table. I wasn't sure of the best way to get prices for the added tokens, so this implementation uses the `sBTC` price for renBTC/wBTC and the `ETH` price for wETH.

Also, in order to accurately show the percentage, I had to add the new token values to the total portfolio value. The result is that the total value of assets in the portfolio table won't be equal to the "Total Synth Value" metric at the top of the page which could be confusing. Here's an example for address `0x1BC64307e6DDC620635b862FD2051BE32Bbf26A3`

image


clementbalestrat (approved):

clementbalestrat:
@mi-yu awesome work, this is such a great PR. Gonna do some testing today.

Are you on Synthetix Discord yet? If yes can you ping me there? It will be easier for us to communicate.

clementbalestrat:
Screen Shot 2021-02-25 at 11 47 07 am
 
@mi-yu  There is probably a glitch with how you calculate the deviation % from the debt pool. sUSD should also be red here.

PR #365, opened 2021-03-04
Feat/layer 2 [NEW BRANCH]

clementbalestrat:


queries/deposits/useGetDepositsDataQuery.ts
```diff
@@ -1,13 +1,13 @@
...
 import { useRecoilValue } from 'recoil';
-import { providers } from 'ethers';
 import orderBy from 'lodash/orderBy';
 import { Watcher } from '@eth-optimism/watcher';
 
 import synthetix from 'lib/synthetix';
 
 import QUERY_KEYS from 'constants/queryKeys';
-import { L1_MESSENGER_ADDRESS, L2_MESSENGER_ADDRESS, OVM_RPC_URL } from 'constants/ovm';
+import { L1_MESSENGER_ADDRESS, L2_MESSENGER_ADDRESS } from 'constants/ovm';
```
justinjmoses, line 9: these can move into the monorepo as another package

justinjmoses (commented):
Suggested a slightly different API

containers/Connector/Connector.tsx
```diff
@@ -39,18 +44,20 @@ const useConnector = () => {
 			const networkId = await getDefaultNetworkId();
 
 			// @ts-ignore
-			const provider = new ethers.providers.InfuraProvider(
+			const providers = defaultProviders({
```
justinjmoses, line 47: How about a slightly different API

```js
import { loadProviders } from `@synthetixio/providers`;

const providers = loadProviders({ infuraID, ... });
const provider = providers.getProviderByNetworkID(window.web3.networkId || networkId);
// ...
// and the provider you return can wrap a regular InfuraProvider with extra props:
console.log('This provider is layer', provider.layer);
// and you can abstract away send/sign logic if you like, so you don't care if it's signs or sends a txn
await provider.submitTransaction(payload); 
// ...
```

PR #399, opened 2021-03-14
fix: NaN under SNX TVL on /earn

JChiaramonte7:
directly querying subgraph for holders

same as this https://github.com/Synthetixio/stats/commit/b1171bf3f1022d1fb38cc021ef55ab05493a835a

### DIFF
--- a/queries/staking/useSNXLockedValueQuery.ts
+++ b/queries/staking/useSNXLockedValueQuery.ts
@@ -7,6 +7,7 @@ import QUERY_KEYS from 'constants/queryKeys';
 
 import { networkState } from 'store/wallet';
 import synthetix from 'lib/synthetix';
+import { getHolders } from './helpers';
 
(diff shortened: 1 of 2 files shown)


queries/staking/useSNXLockedValueQuery.ts
```diff
@@ -46,17 +47,19 @@ const useSNXLockedValueQuery = (options?: QueryConfig<number>) => {
 			let snxLocked = 0;
 
 			for (const { collateral, debtEntryAtIndex, initialDebtOwnership } of holders) {
+				const formattedCollateral = Number(synthetix.js?.utils.formatEther(collateral));
```
evgenyboxer, line 50: could we have a different name for `formattedCollateral` because it hints it might be formatted like a string (e.g - `1000` becomes `1,000.00`)

evgenyboxer (approved):

drptbl (approved):
Tested, it fixes issues mentioned on Discord.

PR #401, opened 2021-03-16
EOA wallet check for L2 deposit & migrate

clementbalestrat:


containers/Connector/Connector.tsx
```diff
@@ -26,8 +31,9 @@ const useConnector = () => {
 	const [onboard, setOnboard] = useState<ReturnType<typeof initOnboard> | null>(null);
 	const [notify, setNotify] = useState<ReturnType<typeof initNotify> | null>(null);
 	const [isAppReady, setAppReady] = useRecoilState(appReadyState);
-	const setWalletAddress = useSetRecoilState(walletAddressState);
+	const [walletAddress, setWalletAddress] = useRecoilState(walletAddressState);
 	const [walletWatched, setWalletWatched] = useRecoilState(walletWatchedState);
+	const [isEOAWallet, setIsEOAWallet] = useRecoilState(isEOAWalletState);
```
evgenyboxer, line 36: Since `isEOAWallet` is not used here, you may want to use `useSetRecoilState` instead.
`const setIsEOAWallet = useSetRecoilState(isEOAWalletState)`

when `useRecoilState` is used, it would subscribe to any changes of `isEOAWallet`, so in those cases, you just want `useSetRecoilState`.

evgenyboxer (approved):
@clementbalestrat lgtm (just one minor change to be addressed)

containers/Connector/Connector.tsx
clementbalestrat: good point

PR #418, opened 2021-03-23
use checksum everywhere instead of lowercase (branched from master)

andytcf:
Fixes the issue where the update in the snapshot strategy does not match ```.toLowerCase()``` to addresses anymore and instead uses the ethers util ```ethers.utils.getAddress() // checksum check``` instead that is more reliable and consistent

### DIFF
--- a/mutations/gov/useSignMessage.ts
+++ b/mutations/gov/useSignMessage.ts
@@ -1,12 +1,12 @@
 import { useRecoilValue } from 'recoil';
 import { useMutation, useQueryCache } from 'react-query';
-// import { version } from '@snapshot-labs/snapshot.js/package.json';
 import axios from 'axios';
 
 import { MESSAGE_URL, SPACE_KEY } from 'constants/snapshot';
 import { walletAddressState } from 'store/wallet';
 import Connector from 'containers/Connector';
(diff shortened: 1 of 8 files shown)


mutations/gov/useSignMessage.ts
```diff
@@ -53,7 +53,7 @@ const useSignMessage = () => {
 		async (payload: SignaturePayload) => {
 			const version = '0.1.3';
 			let msg: any = {
-				address: walletAddress?.toLowerCase(),
+				address: ethers.utils.getAddress(walletAddress ?? ''),
```
clementbalestrat, line 56: can you confirm `ethers.utils.getAddress('')` is not going to return an error?

clementbalestrat (changes requested):
LGTM Candy. Just a minor comment about using `getAddress()` with an empty string. Also History.tsx has a warning with a missing React Hook. Please make sure to run `npm run lint` and see if everything is fine.

Also remove the `package-lock.json` changes.

mutations/gov/useSignMessage.ts
andytcf: @clementbalestrat good pick up

andytcf:
@clementbalestrat actioned those changes requested

PR #433, opened 2021-03-24
L2 fix

clementbalestrat:


containers/Connector/Connector.tsx
```diff
@@ -78,37 +78,51 @@ const useConnector = () => {
...
 							signer,
 						});
-						onboard.config({ networkId });
 
 						if (transactionNotifier) {
 							transactionNotifier.setProvider(provider);
 						} else {
 							setTransactionNotifier(new TransactionNotifier(provider));
 						}
-						setProvider(provider);
```
andytcf, line 101: why don't we set a provider to the recoil state like is done in L118

andytcf (approved):

containers/Connector/Connector.tsx
andytcf: nvm ignore this

PR #448, opened 2021-03-30
Add sETH incentives and disable 0% apys

andytcf:
This feature branch adds the sETH shorting rewards to incentives "/earn".

Others
- Fixes the reference of the lpData via the index, made it reference instead the object so that adding more incentives would not mess up the index of the array.
- Disable 0% apy on the home layouts

### DIFF
--- a/queries/shorts/useSETHShortsQuery.ts
+++ b/queries/shorts/useSETHShortsQuery.ts
@@ -4,7 +4,12 @@ import { useRecoilValue } from 'recoil';
 import synthetix from 'lib/synthetix';
 import QUERY_KEYS from 'constants/queryKeys';
 import { appReadyState } from 'store/app';
-import { walletAddressState, isWalletConnectedState, networkState } from 'store/wallet';
+import {
+	walletAddressState,
+	isWalletConnectedState,
+	networkState,
+	isMainnetState,
+} from 'store/wallet';
 import { Synths } from 'constants/currency';
 import { ShortRewardsData } from './types';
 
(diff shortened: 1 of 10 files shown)


queries/shorts/useSETHShortsQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+import { walletAddressState, isWalletConnectedState, networkState } from 'store/wallet';
+import { Synths } from 'constants/currency';
+import { ShortRewardsData } from './types';
+
+const useSETHShortsQuery = (options?: QueryConfig<ShortRewardsData>) => {
+	const isAppReady = useRecoilValue(appReadyState);
+	const isWalletConnected = useRecoilValue(isWalletConnectedState);
+	const walletAddress = useRecoilValue(walletAddressState);
+	const network = useRecoilValue(networkState);
+
```
clementbalestrat, line 16: Please use `const isMainnet = useRecoilValue(isMainnetState);` and only do the query for Mainnet to avoid any console error on testnet or L2. (See any other pool query file for an example).

queries/shorts/useSETHShortsQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+				openInterestUSD: openInterestUSD,
+				distribution: distribution,
+				periodFinish: Number(periodFinish) * 1000,
+				duration: Number(duration) * 1000,
+				rewards,
+				staked,
+			};
+		},
+		{
+			enabled: isAppReady && isWalletConnected,
```
clementbalestrat, line 69: `enabled: isAppReady && isWalletConnected && isMainnet`

clementbalestrat (changes requested):
LGTM except my comment with the `isMainnet` flag. Also please make sure to check the console for testnet and L2 for any errors.

queries/shorts/useSETHShortsQuery.ts
andytcf: thanks @clementbalestrat

PR #449, opened 2021-03-30
Fix/staking l2 testing

clementbalestrat:


sections/layer2/deposit/DepositTab/TabContent.tsx
```diff
@@ -54,7 +54,7 @@ const TabContent: FC<TabContentProps> = ({
 	const currencyKey = CryptoCurrency['SNX'];
 
 	const renderButton = () => {
-		if (depositAmount) {
+		if (depositAmount && !depositAmount.isZero()) {
```
JChiaramonte7, line 57: I think `depositAmount.toNumber()` could be a more concise way of providing this condition

JChiaramonte7 (approved):

PR #460, opened 2021-04-01
fix: proper L2 urls for block explorer and icons  (#459) release

JChiaramonte7:
* fix: updated optimism-networks with support for proper block explorer and icon urls

* fix: lockfile

* fix: new version with proper build

* fix: new version with https


justinjmoses (approved):

vbstreetz (commented):
Looking good 👍🏽

PR #462, opened 2021-04-07
filter council nominations to exclude invalid proposals

andytcf:
Since the council elections depend on a json that maps their addresses to their discord identity, there were some cases where if a invalid election was created, the proposals would not render correctly.

### DIFF
--- a/queries/gov/useProposals.tsx
+++ b/queries/gov/useProposals.tsx
@@ -13,6 +13,7 @@ import snapshot from '@snapshot-labs/snapshot.js';
 import Connector from 'containers/Connector';
 import { ethers } from 'ethers';
 import CouncilDilution from 'contracts/councilDilution.js';
+import CouncilNominations from 'constants/nominations.json';
 
 const useProposals = (spaceKey: SPACE_KEY, options?: QueryConfig<Proposal[]>) => {
 	const isAppReady = useRecoilValue(appReadyState);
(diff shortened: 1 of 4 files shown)


queries/gov/useProposals.tsx
```diff
@@ -96,6 +97,57 @@ const useProposals = (spaceKey: SPACE_KEY, options?: QueryConfig<Proposal[]>) =>
...
+						return {
+							...proposal,
+							votes: voteCount,
+						};
+					} else {
+						return null;
+					}
+				});
+				const resolvedProposals = await Promise.all(mappedProposals);
+				return resolvedProposals.filter((e) => e !== null);
```
evgenyboxer, line 150: are you looking to filter just `null` or also `undefined`? because `e != null` would match both (which is sometimes really useful)

sections/gov/components/Proposal/Results.tsx
```diff
@@ -30,12 +30,14 @@ const Results: React.FC<ResultsProps> = ({ hash }) => {
 				const currentElectionMembers = CouncilNominations as any;
 				const mappedProfiles = [] as any;
 
-				currentElectionMembers[electionCount].forEach((member: any) => {
-					mappedProfiles.push({
-						address: member.address,
-						name: member.discord,
+				if (currentElectionMembers[hash]) {
```
evgenyboxer, line 33: seems like the react hook `useEffect` is complaining that its missing `hash` (it shows on GitHub)

evgenyboxer (approved):

queries/gov/useProposals.tsx
andytcf: both is good, but mainly null

sections/gov/components/Proposal/Results.tsx
andytcf: thanks @evgenyboxer

PR #475, opened 2021-04-12
browser wallet added

clementbalestrat:
This PR:

- Adds browser wallet support (any wallet which injects window.ethereum) such as Metamask, Frame.sh...etc
- Fixes the Input colour when selecting a wallet address for Ledger & Trezor.
- Updates the "please-install-metamask" copy


andytcf (approved):
LGTM

andytcf:
@drptbl only workflow that seems to be broken from my quick QA is connecting directly using a ledger nano x

drptbl:
@clementbalestrat e2e tests were failing because of these changes, I've updated them for you.

evgenyboxer (approved):
lgtm!

PR #477, opened 2021-04-13
Show ETH/BTC balance even if user has no synths

mi-yu:
Currently, a user's debt portfolio table will not have an entry for ETH-related assets if they do not have any sETH. This is also true for BTC-related assets and sBTC. This can be confusing since the assets presented would not add up to 100%. For example:

Currently:
image

This PR:
image

### DIFF
--- a/sections/debt/components/PortfolioTable/PortfolioTable.tsx
+++ b/sections/debt/components/PortfolioTable/PortfolioTable.tsx
@@ -29,8 +29,8 @@ import Info from 'assets/svg/app/info.svg';
 const SHOW_HEDGING_INDICATOR_THRESHOLD = new BigNumber(0.1);
 
 type DebtPoolTableProps = {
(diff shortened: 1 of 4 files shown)


sections/debt/components/PortfolioTable/PortfolioTable.tsx
```diff
@@ -29,8 +29,8 @@ import Info from 'assets/svg/app/info.svg';
 const SHOW_HEDGING_INDICATOR_THRESHOLD = new BigNumber(0.1);
 
 type DebtPoolTableProps = {
-	synthBalances: any;
-	cryptoBalances: any;
+	synthBalances: CryptoBalance[];
```
clementbalestrat, line 32: nice

clementbalestrat (approved):
Great job @mi-yu !

PR #481, opened 2021-04-14
Debt pool vs sUSD chart

mi-yu:
Only includes debt chart for now

image


hooks/useGlobalHistoricalDebtData.ts
```diff
@@ -0,0 +1,68 @@
...
+			eventBlocks.forEach((event, i) => {
+				const multiplier = event.type === StakingTransactionType.Burned ? -1 : 1;
+				const aggregation =
+					data.length === 0
+						? multiplier * event.value
+						: multiplier * event.value + data[i - 1].issuance;
+				if (data.length === 0) {
+					debtPoolStartingValue = event.totalDebt;
+				}
+
```
clementbalestrat, line 49: Logic LGTM but I've got a couple of comments. I'm curious to see what you think:

- So because you'll get the debt pool value from let's say a year ago, and will start your aggregation from 0 (or at least with the first burn/mint event value) I was thinking we could "hack" the data and set `issuance` to be equal to `debtPoolStartingValue`. Then both lines will start from the same point and we will see diverge at some point in `t + n`. Thought?

- Also I'm not sure what's the logic behind `debtPool: event.totalDebt - debtPoolStartingValue`. Are you subtracting  `debtPoolStartingValue` from `event.totalDebt` every time?

mi-yu: - The first point makes sense, I can try that!
- I wanted to normalize all of the `totalDebt` values to whatever the initial value was at the start of the time range, so the graph would actually show the change in total debt starting from the earliest data point. The goal was to put `totalDebt` values on the same scale as issued/burned so the comparison made more sense, but I'm not sure if this actually accomplishes that

hooks/useHistoricalDebtData.ts
```diff
@@ -17,7 +17,7 @@ type HistoricalDebtAndIssuance = {
 	data: HistoricalDebtAndIssuanceData[] | [];
 };
 
-const useHistoricalDebtData = () => {
+const useHistoricalDebtData = (global = false) => {
```
clementbalestrat, line 20: Is `global` needed here?

package.json
```diff
@@ -61,7 +61,7 @@
 		"styled-components": "5.1.1",
 		"styled-media-query": "2.1.2",
 		"svg-react-loader": "0.4.6",
-		"synthetix-data": "2.1.42",
+		"synthetix-data": "^2.1.43",
```
clementbalestrat, line 64: let's lock the version `"synthetix-data": "2.1.43"`

queries/staking/useDailyBurnedQuery.tsx
```diff
@@ -0,0 +1,29 @@
...
+const useDailyBurnedQuery = (options?: QueryConfig<DailyStakingRecord[]>) => {
+	const isWalletConnected = useRecoilValue(isWalletConnectedState);
+	const network = useRecoilValue(networkState);
+
+	return useQuery<DailyStakingRecord[]>(
+		QUERY_KEYS.Staking.Burned('', network?.id!),
+		async () => {
+			const now = new Date();
+			const minTimestamp = now.getTime() / 1000 - 180 * 86400;
+			const dailyBurneds = await snxData.snx.dailyBurned({ max: 6 * 30, minTimestamp });
```
clementbalestrat, line 19: Any reason why max is set to `180` ?

queries/staking/useDailyIssuedQuery.tsx
```diff
@@ -0,0 +1,29 @@
...
+const useDailyIssuedQuery = (options?: QueryConfig<DailyStakingRecord[]>) => {
+	const isWalletConnected = useRecoilValue(isWalletConnectedState);
+	const network = useRecoilValue(networkState);
+
+	return useQuery<DailyStakingRecord[]>(
+		QUERY_KEYS.Staking.Issued('', network?.id!),
+		async () => {
+			const now = new Date();
+			const minTimestamp = now.getTime() / 1000 - 180 * 86400;
+			const dailyIssueds = await snxData.snx.dailyIssued({ max: 6 * 30, minTimestamp });
```
clementbalestrat, line 19: Same here

sections/debt/Debt.tsx
```diff
@@ -1,26 +1,31 @@
...
 
 import PorfolioTable from './components/PortfolioTable';
 import useCryptoBalances from 'hooks/useCryptoBalances';
+import OverviewTab from './components/OverviewTab';
+import ManageTab from './components/ManageTab';
+import DebtTabs from './components/DebtTabs';
 
 const DebtSection = () => {
 	const { t } = useTranslation();
+	const router = useRouter();
```
clementbalestrat, line 28: Unused variable apparently

sections/debt/components/DebtHedgingInfoPanel/DebtHedgingChart.tsx
```diff
@@ -0,0 +1,97 @@
...
+import colors from 'styles/theme/colors';
+import fonts from 'styles/theme/fonts';
+
+import SpinnerIcon from 'assets/svg/app/loader.svg';
+import ChartLabel from './ChartLabel';
+
+const DebtHedgingChart: React.FC = () => {
+	const { t } = useTranslation();
+	const { data, isLoading } = useGlobalHistoricalDebtData();
+	// console.log(data);
```
clementbalestrat, line 20: let's remove this `console.log`

clementbalestrat (changes requested):
Awesome work @mi-yu ! Some minors comments before we can merge it

clementbalestrat (approved):

clementbalestrat (approved):

PR #484, opened 2021-04-15
Use last available value for issued debt if missing at index

mi-yu:
Fixes issue like this:
image

With change:
image


mi-yu:
fixed by https://github.com/Synthetixio/staking/pull/485

Issue #455, opened 2021-03-31
Make it more clear for users that they can't use other wallets than metamask for L2


drptbl:
**Description:**
Make it more clear for users that they can't use other wallets than metamask for L2. Options related to L2 should be dimmed out in case someone uses different wallet method than metamask with some information that metamask is the only wallet supported for L2 on "Layer 2" page.

For example here:
<img width="1104" alt="Screenshot 2021-03-31 at 18 07 08" src="https://user-images.githubusercontent.com/8177587/113175751-0ab90700-924c-11eb-81ee-5b93f2bc1f14.png">

Boda805:
I'd like to take an attempt at this if no-one else minds.

JChiaramonte7:
@drptbl is what we have now sufficient? it doesn't block users without a metamask wallet, but there's a pretty strong disclaimer

![image](https://user-images.githubusercontent.com/4015977/116474609-b9019d80-a846-11eb-9aca-55366d160f86.png)

PR #492, opened 2021-04-16
#455 Make clear only MetaMask L2

Boda805:
Styling of text on L2 page could be better but this should do the trick


clementbalestrat:
@Boda805 can you please describe what your PR does? I had a look at the code and I'm a bit confused on what you are trying to achieve here?

Boda805:
It's to answer issue #455 to dim out options for L2 when a wallet other than MetaMask is selected, as it's stated on the issue that MetaMask is only wallet supported for L2. I probably should have done a draft as some of it is a little crude yet but let me know what you think.

clementbalestrat:
@Boda805 Thanks! I'll try to have a look at it today or tomorrow.

JChiaramonte7 (changes requested):
Hey @Boda805 thanks for the PR. Currently discussing with @drptbl on #455, but I don't think we need this because we already have a pretty strong disclaimer. 

Also could you regenerate your lockfile with npm 6? We haven't switched to 7 yet, but we'll probably do this in a separate PR.

clementbalestrat:
Closing that one for now until we decide what to do.

PR #503, opened 2021-04-26
Refactor Staking dApp 

andytcf:
This PR aims to:
- Convert currency icon to use tokenlists and synthetix/assets as a fallback so that we can automatically get the latest currency assets deployed
- Adds the balancer incentives to the frontpage layout
- Disables 1inch queries on L2
- Fixes the progress bar when the timer has ended on incentives (makes it full rainbow width)

### DIFF
--- a/constants/currency.ts
+++ b/constants/currency.ts
@@ -89,4 +89,6 @@ export const FIAT_SYNTHS = new Set([
 	Synths.sCHF,
 ]);
 
+export const ETH_ADDRESS = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
+
 export const sUSD_EXCHANGE_RATE = 1;
--- a/sections/earn/IncentivesDefault.tsx
+++ b/sections/earn/IncentivesDefault.tsx
@@ -70,6 +70,7 @@ const Incentives: FC<IncentivesProps> = ({
 							staked: {
 								balance: stakedAmount,
 								asset: CryptoCurrency.SNX,
+								ticker: CryptoCurrency.SNX,
 							},
 							rewards: stakingRewards.toNumber(),
 							periodStarted: currentFeePeriodStarted.getTime(),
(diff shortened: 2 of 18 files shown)


constants/currency.ts
```diff
@@ -89,4 +89,6 @@ export const FIAT_SYNTHS = new Set([
 	Synths.sCHF,
 ]);
 
+export const ETH_ADDRESS = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
```
dvd-schwrtz, line 92: maybe make note it is just a random eth address or a placeholder eth address instead of just naming it eth address?

dvd-schwrtz (approved):

constants/currency.ts
JChiaramonte7: I think this is the ETH address that 1inch uses, but yeah maybe `ONEINCH_ETH_ADDRESS` or something

JChiaramonte7 (approved):
Much needed fixes 🙌

sections/earn/IncentivesDefault.tsx
```diff
@@ -70,6 +70,7 @@ const Incentives: FC<IncentivesProps> = ({
 							staked: {
 								balance: stakedAmount,
 								asset: CryptoCurrency.SNX,
+								ticker: CryptoCurrency.SNX,
```
clementbalestrat, line 73: can `ticker` and `asset` be different?

sections/earn/IncentivesMainnet.tsx
```diff
@@ -376,24 +375,29 @@ const Incentives: FC<IncentivesProps> = ({
 						needsToSettle={lpData[Synths.iETH].data?.needsToSettle}
 					/>
 				)}
-				{activeTab === Tab.sTLSA_LP && (
+				{/* {activeTab === Tab.sTLSA_LP && (
```
clementbalestrat, line 378: Do we still need this?

clementbalestrat (commented):
LGTM, just a few questions.

PR #504, opened 2021-04-26
feat: estimateGas on L2

JChiaramonte7:
## Enable L2 Gas

The gasLimit on L2 is an [encoded](notion.so/How-to-pay-Fees-in-Optimistic-Ethereum-f706f4e5b13e460fa5671af48ce9a695) value that carries information about fee cost on L1 + gas limit on L2. We should get the gasLimit directly from `estimateGas`. This PR removes the buffer typically added from `normalizeGasLimit` on gasLimit calculations on L2.

resolves https://github.com/Synthetixio/issues/issues/42


dvd-schwrtz (approved):

vbstreetz:
Very interesting gas price 😆

Screenshot 2021-06-08 at 20 40 30

vbstreetz (approved):
Looking good 👍🏽

PR #505, opened 2021-04-28
Same data source for percentage in side nav and asset table.

crypto-supporter:
Didnt have any env variables so havent tested this.
Feels like line chart is also displaying wrong data.

### DIFF
--- a/sections/shared/Layout/SideNav/PriceItem.tsx
+++ b/sections/shared/Layout/SideNav/PriceItem.tsx
@@ -1,5 +1,5 @@
+import React, { FC } from 'react';
 import styled from 'styled-components';
-import { FC } from 'react';
 import { Trans } from 'react-i18next';
 
 import { FlexDivCentered, FlexDivCol, FlexDivRowCentered, NoTextTransform } from 'styles/common';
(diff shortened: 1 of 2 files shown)


sections/shared/Layout/SideNav/PriceItem.tsx
```diff
@@ -93,32 +98,9 @@ const PriceInfo = styled(FlexDivCentered)`
 	justify-content: space-between;
 `;
 
-const PercentChange = styled.div<{ trendLinePositive: boolean }>`
+const PercentChange = styled.div`
 	font-size: 10px;
 	font-family: ${(props) => props.theme.fonts.interBold};
-	color: ${(props) =>
```
andytcf, line 99: see comment above about reintroducing the correct colors

```
green is up
pink is down
```

andytcf (changes requested):

sections/shared/Layout/SideNav/PriceItem.tsx
crypto-supporter: @andytcf Isnt that already handled in `ChangePercent.tsx`?
image

The only difference between the percentage design in the AssetTable from the design in the side nav is the font-family and font-size.

I usually use css modules, havent worked with styled-components before so maybe i dont understand the how things styles are inherited to child doms. 


You mean something like this:

```
<PercentChange isPositive={historicalRates.data?.change >= 0}>
      <ChangePercent className="percent" value={historicalRates.data?.change} />
</PercentChange>

const PercentChange = styled.div<{ isPositive: boolean }>`
	font-size: 10px;
	font-family: ${(props) => props.theme.fonts.interBold};
	color: ${(props) => (props.isPositive ? props.theme.colors.green : props.theme.colors.pink)};
`;
```

However it feels redundant to set color here when it is already handled by the `ChangePercent` component?

clementbalestrat:
Closing this one since no fixes have been pushed since last review.

PR #512, opened 2021-04-29
Feat/token transfer

clementbalestrat:
Screen Shot 2021-04-29 at 8 16 55 pm


JChiaramonte7 (changes requested):
Everything else looks good. I just found one bug trying to transfer SNX on L2. It doesn't look like my balance is being detected on the transfer modal. It does show up in the assets table, however.

image
image

dvd-schwrtz (approved):

JChiaramonte7 (approved):

PR #520, opened 2021-04-29
fix: Issues with Staking Earn

JChiaramonte7:


components/Table/Table.tsx
```diff
@@ -14,7 +14,7 @@ import Pagination from './Pagination';
 export type TablePalette = 'primary';
 
 const CARD_HEIGHT = '40px';
-const MAX_PAGE_ROWS = 10;
+const MAX_PAGE_ROWS = 12;
```
JChiaramonte7, line 17: There is a UI bug where react table pagination doesn't play nice with how we remove columns from the table if there's an `activeTab`. Would like to resolve this issue eventually, but going to increase max page rows for now.

sections/earn/IncentivesTable.tsx
```diff
@@ -187,7 +187,7 @@ const IncentivesTable: FC<IncentivesTableProps> = ({ data, isLoaded, activeTab }
 				Header: <>{t('earn.incentives.options.rewards.title')}</>,
 				accessor: 'rewards',
 				Cell: (cellProps: CellProps<EarnItem, EarnItem['rewards']>) => {
-					const isDualRewards = cellProps.row.original.staked.asset === LP.UNISWAP_DHT;
+					const isDualRewards = typeof cellProps.row.original.rewards === 'object';
```
dvd-schwrtz, line 190: this doesn't feel like a good solution. can we do better?

components/Table/Table.tsx
dvd-schwrtz: hmm... ok, let's add this to the list of issues to look into

sections/earn/IncentivesTable.tsx
JChiaramonte7: yeah good point, fixed with something better

components/Table/Table.tsx
JChiaramonte7: #521

clementbalestrat (approved):

clementbalestrat:
LGTM @JChiaramonte7 can you please remove `package-lock.json` from your commit?

components/Table/Table.tsx
justinjmoses: why even use pagination when we have scrolling?

justinjmoses (approved):

PR #523, opened 2021-05-01
fix: Clear Debt

JChiaramonte7:
- adds error state for not enough ETH in wallet
- also changes burn amount to === all sUSD in wallet (which is debtBalance + 1inch buffer) to account for possible debt shifts


clementbalestrat (approved):
LGTM

andytcf (approved):

PR #545, opened 2021-05-20
Feat/snapshot graphql

andytcf:
This PR replaces the axios.fetch with graphql endpoints for the governance section

This PR also has some improvements to the gov:
- Refactor the hook to get the active proposals to a graphql custom query
- Only fetch the proposals if a user has clicked on the relevant space tab (i.e snxgov.eth) ~ lazy loading
- Gets the latest council election in a custom graphql query, more efficient as I won't be pulling down the entire councilProposals only to map through it @bonustrack nice work here
- Virtualized List for the voting history to improve loading times
- Refactor the proposal so it only needs to grab it once and passes it as a prop 
- Passes in the current results + history as a prop in one parent element to avoid double graphql calls


clementbalestrat (approved):
LGTM candy

queries/gov/useActiveProposalsQuery.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { appReadyState } from 'store/app';
+import { Proposal } from './types';
+import { networkState, walletAddressState } from 'store/wallet';
+import request, { gql } from 'graphql-request';
+
+const useActiveProposalsQuery = (options?: QueryConfig<number>) => {
+	const isAppReady = useRecoilValue(appReadyState);
+	const network = useRecoilValue(networkState);
+	const walletAddress = useRecoilValue(walletAddressState);
+	const isL1 = !network?.useOvm ?? true;
```
vbstreetz, line 16: i'd use the isL2State helper..

queries/gov/useLatestCouncilElectionQuery.ts
```diff
@@ -0,0 +1,61 @@
...
+import { networkState, walletAddressState } from 'store/wallet';
+import request, { gql } from 'graphql-request';
+import { Proposal } from './types';
+
+const useLatestCouncilElectionQuery = (options?: QueryConfig<Proposal>) => {
+	const isAppReady = useRecoilValue(appReadyState);
+	const walletAddress = useRecoilValue(walletAddressState);
+	const network = useRecoilValue(networkState);
+	const isL1 = !network?.useOvm ?? true;
+
```
vbstreetz, line 15: same

vbstreetz (changes requested):
Looking good 👍

PR #552, opened 2021-05-25
Feat/liquidation banner

clementbalestrat:
Screen Shot 2021-05-25 at 4 48 56 pm

This PR will display the liquidation banner as a priority. I don't think we've got a design good enough to have multiple banners at the moment. Hence the `if` statement in `BannerManager.tsx`.

This code could definitely be better at some point, to handle multiple banners at the same time.

### DIFF
--- a/components/BannerManager/BannerManager.tsx
+++ b/components/BannerManager/BannerManager.tsx
@@ -22,7 +22,7 @@ const BannerManager: FC = () => {
 		? 0
 		: 100 / Number(ethersUtils.formatEther(liquidationRatio.toString()));
 
-	if (!liquidationDeadlineForAccount?.isZero()) {
+	if (!liquidationDeadlineForAccount.isZero()) {
 		return (
 			<Banner
 				type={BannerType.WARNING}
--- a/components/BannerManager/BannerManager.tsx
+++ b/components/BannerManager/BannerManager.tsx
@@ -1,0 +1,75 @@
+import { FC } from 'react';
+import { Trans } from 'react-i18next';
+import styled from 'styled-components';
+import { utils as ethersUtils } from 'ethers';
+
+import useGetLiquidationData from 'queries/liquidations/useGetLiquidationDataQuery';
+
+import Banner, { BannerType } from 'sections/shared/Layout/Banner';
+import { LOCAL_STORAGE_KEYS } from 'constants/storage';
+import { ExternalLink } from 'styles/common';
+import { zeroBN } from 'utils/formatters/number';
+import { formatShortDateWithTime } from 'utils/formatters/date';
+
+const BannerManager: FC = () => {
+	const liquidationData = useGetLiquidationData();
+
+	const liquidationRatio = liquidationData?.data?.liquidationRatio ?? zeroBN;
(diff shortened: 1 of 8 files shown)


components/BannerManager/BannerManager.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+	const liquidationRatio = liquidationData?.data?.liquidationRatio ?? zeroBN;
+	const liquidationDeadlineForAccount =
+		liquidationData?.data?.liquidationDeadlineForAccount ?? zeroBN;
+
+	const liquidationRatioPercentage = liquidationRatio.isZero()
+		? 0
+		: 100 / Number(ethersUtils.formatEther(liquidationRatio.toString()));
+
+	if (!liquidationDeadlineForAccount?.isZero()) {
```
andytcf, line 25: i think we can remove the optional ```?``` here since the 

```
const liquidationDeadlineForAccount
```

always returns a value

components/BannerManager/BannerManager.tsx
```diff
@@ -0,0 +1,75 @@
...
+		);
+	} else {
+		return (
+			<Banner
+				type={BannerType.ATTENTION}
+				localStorageKey={LOCAL_STORAGE_KEYS.WARNING_URL_BANNER_VISIBLE}
+				message={
+					<Trans
+						i18nKey={'user-menu.banner.url-warning'}
+						components={[<StyledExternalLink href="https://staking.synthetix.io" />]}
```
andytcf, line 55: would this not be a simple route to ```/```

sections/shared/Layout/Banner/Banner.tsx
```diff
@@ -1,20 +1,31 @@
...
+	type?: BannerType;
 };
 
-const Banner: FC<BannerProps> = ({ message, localStorageKey }) => {
+const Banner: FC<BannerProps> = ({ message, localStorageKey, type = BannerType.INFORMATION }) => {
 	const [isBannerVisible, setIsBannerVisible] = useState<boolean>(true);
 
 	const fetchFromLocalStorage = useCallback(() => {
+		if (!localStorageKey) {
+			setIsBannerVisible(true);
```
andytcf, line 26: is this meant to be ```false``` since localStorageKey determines which banner to show?

andytcf (changes requested):

components/BannerManager/BannerManager.tsx
clementbalestrat: Kinda the same tho.

sections/shared/Layout/Banner/Banner.tsx
clementbalestrat: `WARNING` can't be dismissed. That's something I decided. I don't want people to hide their liquidation warnings and then forget about it.

So if there is no storage key, then you assume you can't hide the banner.

components/BannerManager/BannerManager.tsx
clementbalestrat: true dat

vbstreetz (approved):
Looking good..

PR #562, opened 2021-06-01
add a fallback for zapper list

andytcf:


queries/tokenLists/useZapperTokenList.ts
```diff
@@ -24,7 +24,7 @@ const useZapperTokenList = (options?: QueryConfig<TokenListQueryResponse>) => {
 	return useQuery<TokenListQueryResponse>(
 		QUERY_KEYS.TokenLists.Zapper,
 		async () => {
-			const response = await axios.get<TokenListResponse>(EXTERNAL_LINKS.TokenLists.Zapper);
+			const response = await axios.get<TokenListResponse>(EXTERNAL_LINKS.TokenLists.OneInch);
```
vbstreetz, line 27: Should this remain as is?

vbstreetz (changes requested):
Added a comment but looking good.

queries/tokenLists/useZapperTokenList.ts
andytcf: ah yes it should

vbstreetz (approved):
looks good 👍🏽

PR #600, opened 2021-06-26
Changes to API calls, locked the claiming when user has not voted, UI fixes

andytcf:
This pull request features
- Locks the claiming button when a user has not voted for all three DAO elections yet
- Removes the vote count when getting all proposals in the table to prevent api call limit
- Adds the highlighted frontend for gdao (5 members) adao (3 members)
- Refactors alot of queries


vbstreetz:
Might need some responsive ui fixes..

PR #615, opened 2021-07-07
[Feature]: Add a selectable buffer when trying to burn max debt on Staking dApp

andytcf:
Fixes https://github.com/synthetixio/issues/issues/190

To test
1) Mint max
2) Try to burnMax (input should be blocked if you choose a slippage > 0, unless you have spare sUSD)
3) Acquire more sUSD
4) Try to burn with buffer, if successful it should burn only the debt YOU have and not more

Other notes
- Should allow for 0% buffer burning max

image

image

### DIFF
--- a/components/BufferSelector/BufferSelector.tsx
+++ b/components/BufferSelector/BufferSelector.tsx
@@ -1,0 +1,183 @@
+import React, { useMemo, useState } from 'react';
+import styled from 'styled-components';
+import Tippy from '@tippyjs/react';
+import { useTranslation } from 'react-i18next';
+import { Svg } from 'react-optimized-image';
+
+import InfoIcon from 'assets/svg/app/info.svg';
+
+import Button from 'components/Button';
+
+import { NumericValue, FlexDivRowCentered, FlexDivRow } from 'styles/common';
+
+import { formatPercent } from 'utils/formatters/number';
+import NumericInput from 'components/Input/NumericInput';
+
+type BufferSelectorProps = {
+	setBuffer: (num: string) => void;
+	buffer: string;
+};
+
+const BufferSelector: React.FC<BufferSelectorProps> = ({ buffer, setBuffer, ...rest }) => {
+	const { t } = useTranslation();
+
+	const [customBuffer, setCustomBuffer] = useState<string>('');
+
+	const BUFFER_VALUES = useMemo(
+		() => [
+			{
+				text: t('common.buffer.low'),
+				value: '0.0010',
+			},
+			{
+				text: t('common.buffer.medium'),
+				value: '0.0025',
+			},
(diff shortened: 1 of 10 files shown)


components/BufferSelector/BufferSelector.tsx
```diff
@@ -0,0 +1,186 @@
...
+	padding: 16px 0 8px 0;
+`;
+
+const BufferHelperTooltip = styled(Tippy)`
+	background: ${(props) => props.theme.colors.navy};
+	border: 0.5px solid ${(props) => props.theme.colors.navy};
+	border-radius: 4px;
+	width: 120px;
+	.tippy-content {
+		padding: 0;
```
JChiaramonte7, line 126: I would add some padding to give the tooltip content from separation from the surrounding content.

JChiaramonte7 (changes requested):
- [x] Mint max
- [x] Try to burnMax (input should be blocked if you choose a slippage > 0, unless you have spare sUSD)
- [x] Acquire more sUSD
- [x] Try to burn with buffer, if successful it should burn only the debt YOU have and not more

I think it would be good maybe to add some sort of `~` to the output amount to express the sUSD remaining is an estimate. ie. In my case I ended up with a lot more USD because the debt pool didn't shift

image

sections/staking/components/StakingInput/StakingInput.tsx
```diff
@@ -327,6 +333,11 @@ const StakingInput: React.FC<StakingInputProps> = ({
 						</RowTitle>
 						<RowValue>{equivalentSNXAmount}</RowValue>
 					</DataRow>
+					{burnType === BurnActionType.MAX && debtBuffer && setDebtBuffer && (
```
clementbalestrat, line 336: do you need to test if `setDebtBuffer` exists here?

clementbalestrat (changes requested):
LGTM. Some small fixes required before merging.

sections/staking/components/StakingInput/StakingInput.tsx
andytcf: yes because mint tab and burn tab both share this staking input, only burnTab will be pass the props ```debtBuffer and setDebtBuffer``` so its undefined in some cases

andytcf:
@JChiaramonte7 @clementbalestrat 

1) Add padding to the tooltip

image

2) Remove the sUSD currencyKey as its not needed (the label says it already)
image

sections/staking/components/StakingInput/StakingInput.tsx
clementbalestrat: awesome

clementbalestrat (approved):

PR #617, opened 2021-07-07
feat(L2): Add link to etherscan to execute withdrawal relays

JChiaramonte7:
Optimism is no longer relaying withdrawals for users. We need to have a link to help the user self execute their withdrawals until a longer term solution is designed.

image


clementbalestrat (approved):

clementbalestrat:
LGTM.

PR #622, opened 2021-07-12
Fix/wrong network error

andytcf:
Fixes https://github.com/synthetixio/issues/issues/239

Flows to test

- [x] When a user is not connected via a wallet AND they are on the wrong network, there should be an error message shown under the "connect wallet" button
- [ ] From the flow above, if they switch their network to a supported one the page should refresh, calling init() again and rehydrating the page, they should be able to "connect wallet" successfully
- [x] When a user is already successfully connected, if they change to another supported network, the page should NOT refresh (aka ETH to OVM or KOVAN)
- [x] When a user is already successfully connected and they change to a non-supported network, the page will refresh and call init() again and show the same error as the first flow. 

I've decided to go with the pattern recommended here https://docs.metamask.io/guide/ethereum-provider.html#chainchanged

```We strongly recommend reloading the page on chain changes, unless you have good reason not to.```

### CODE REVIEW

File: utils/network.ts [hunk 1]
```diff
@@ -35,3 +35,8 @@ export const getTransactionPrice = (
 export const normalizeGasLimit = (gasLimit: number) => gasLimit + DEFAULT_GAS_BUFFER;
 
 export const normalizedGasPrice = (gasPrice: number) => gasPrice * GWEI_UNIT;
+
+export const matchesNetworkErrorString = (error: string) =>
+	error.includes('unsupported network or network id passed');
+
+export const networkErrorMessage = 'Wrong network detected';
```
clementbalestrat, line 42: let's i18n this maybe?

containers/Connector/config.ts
```diff
@@ -52,7 +53,7 @@ export const initOnboard = (network: Network, subscriptions: Subscriptions) => {
 				},
 				{
 					walletName: 'walletConnect',
-					rpc: { [network.id]: infuraRpc },
+					rpc: { [network ? network.id : 1]: infuraRpc },
```
clementbalestrat, line 56: maybe you can use Network.Mainnet instead of `1`

containers/index.tsx
```diff
@@ -11,15 +11,15 @@ type WithAppContainersProps = {
...
-	<Connector.Provider>
-		<BlockExplorer.Provider>
-			<TransactionNotifier.Provider>
-				<Loans.Provider>
-					<SideNav.Provider>{children}</SideNav.Provider>
-				</Loans.Provider>
-			</TransactionNotifier.Provider>
-		</BlockExplorer.Provider>
-	</Connector.Provider>
+	<SideNav.Provider>
```
clementbalestrat, line 14: Where does this change come from?

clementbalestrat (approved):
Looking good. Just a few comments

containers/index.tsx
andytcf: its because the way its structured, the network error message is rendered by the SideNav.Provider and since most of the logic controlling that is connector it needs to be rearranged

PR #643, opened 2021-07-30
fix: renBTC doesn't exist on L2

andytcf:


andytcf:
@liamzebedee closing this PR as its already in dev actioned by mitch

PR #660, opened 2021-09-07
default token icon to synthetix token list

clementbalestrat:


andytcf (approved):
lgtm, wow must have been old code? i think this was fixed for kwenta

andytcf (changes requested):
need to consider the earn pages, as they use icons not on our synthetix token list

andytcf (approved):

PR #665, opened 2021-09-14
feat: accounts merge

vbstreetz:
Implements https://github.com/Synthetixio/issues/issues/323


sections/merge-accounts/burn/BurnActionBox.tsx
```diff
@@ -0,0 +1,333 @@
...
+	// funcs
+
+	const connectOrBurn = async () => {
+		if (!isWalletConnected) {
+			return connectWallet();
+		}
+		burn();
+	};
+
+	const burn = async () => {
```
dbeal-eth, line 176: would encourage you to use the new `useSynthetixTxn` mutative function in `useSynthetixQueries`--basically just put the data from `getBurnTxData` direcly into it, it will give you gas limit estimates and handle a lot of the state and erroring issues automatically for you

sections/merge-accounts/burn/BurnInfoBox.tsx
```diff
@@ -0,0 +1,247 @@
...
+	const synthsBalancesQuery = useSynthsBalancesQuery(walletAddress);
+
+	const amountToBurn = useRecoilValue(amountToBurnState);
+
+	const sUSDBalance = synthsBalancesQuery?.data?.balancesMap[Synths.sUSD]?.balance ?? wei(0);
+
+	const stakingInfo = useMemo(() => {
+		const calculatedTargetBurn = Math.max(debtBalance.sub(issuableSynths).toNumber(), 0);
+
+		let amountToBurnBN = parseSafeWei(amountToBurn, 0);
```
dbeal-eth, line 71: lol I know this is how I left it after refactoring but maybe its time to rename this from `amountToBurnBN`

also why do you use `, 0)` argument here? This specifies the number of decimals, and if you are burning an amount it would be 18 decimals. It should work best without the extra parameter since it defaults to 18.

sections/merge-accounts/common.ts
```diff
@@ -0,0 +1,227 @@
+import styled from 'styled-components';
```
dbeal-eth, line 1: not relevant for this PR necessarily but we should try reusing more of the components globally within the staking app rather than having common components for each

sections/merge-accounts/merge/MergeActionBox.tsx
```diff
@@ -0,0 +1,411 @@
...
+		return () => {
+			unsubs.forEach((unsub) => unsub());
+		};
+	}, [properSourceAccountAddress, isAppReady, synthetixjs]);
+
+	// funcs
+
+	const onEnterAddress = (e: any) => setSourceAccountAddress((e.target.value ?? '').trim());
+
+	const connectOrMerge = async () => {
```
dbeal-eth, line 248: again you can save like half of this code by using `useSynthetixTxn` mutation

sections/merge-accounts/nominate/NominateActionBox.tsx
```diff
@@ -0,0 +1,377 @@
...
+	const onEnterAddress = (e: any) => setDestinationAccountAddress((e.target.value ?? '').trim());
+
+	const connectOrBurnOrNominate = async () => {
+		if (!isWalletConnected) {
+			return connectWallet();
+		}
+		nominate();
+	};
+
+	const nominate = async () => {
```
dbeal-eth, line 203: here also

dbeal-eth (changes requested):
I'm sure functionally its good but can we reuse more components in order to reduce the overall amount of new code? its extremely expensive to maintain code like this which has lots of unique components

sections/merge-accounts/burn/BurnInfoBox.tsx
vbstreetz: looks like the second param is actually just the default value to return if the value is invalid https://github.com/Synthetixio/staking/blob/dev/utils/parse.ts#L3

sections/merge-accounts/burn/BurnActionBox.tsx
vbstreetz: done

sections/merge-accounts/merge/MergeActionBox.tsx
vbstreetz: done

sections/merge-accounts/nominate/NominateActionBox.tsx
vbstreetz: done

PR #676, opened 2021-09-15
feat: update l2 deposit info copy

vbstreetz:


dbeal-eth (approved):
thanks for taking care of this :pray:

PR #683, opened 2021-09-17
fix: l2 claim issue due to gas estimate normalisation

vbstreetz:


dbeal-eth (approved):
odd that this change is necessary, this is a departure from previous functionality and normally having a small gas limit is no trouble for block limits

vbstreetz:
> odd that this change is necessary, this is a departure from previous functionality and normally having a small gas limit is no trouble for block limits

also just disabled here actually https://github.com/Synthetixio/js-monorepo/commit/c1085e277952a37954049cc00f01e36fef5ab5de

PR #685, opened 2021-09-17
remove document matomo

dbeal-eth:
there are 2 matomo script injections in this app. Only one needed is in `_app`.


drptbl:
@dbeal-eth lgtm 🚀

drptbl (approved):

PR #717, opened 2021-09-24
dev => master

vbstreetz:


sections/merge-accounts/burn/BurnActionBox.tsx
```diff
@@ -0,0 +1,240 @@
...
+		SNXRate,
+		currentCRatio,
+	} = useBurnTx();
+
+	const stakeInfo = useCallback(
+		(burnAmount: Wei): Wei => wei(getStakingAmount(targetCRatio, burnAmount, SNXRate)),
+		[SNXRate, targetCRatio]
+	);
+
+	const burnAmount = debtBalance.gt(sUSDBalance) ? wei(sUSDBalance) : debtBalance;
```
clementbalestrat, line 93: @vbstreetz we want to burn max? So let's take the sUSD balance as the burn amount and allow the burn when balance >= debt. Because the way you do it could lead to issue if the debt shifts while the transaction is being mined.

clementbalestrat (approved):

sections/merge-accounts/burn/BurnActionBox.tsx
vbstreetz: @clementbalestrat does this need to be updated as well? https://github.com/Synthetixio/staking/blob/dev/sections/staking/components/BurnTab/BurnTab.tsx#L50-L51

andytcf (approved):

andytcf (approved):

PR #733, opened 2021-10-08
Fix/claim status

0xclem:


bachstatter:
Added a PR with test and the suggested simplification

PR #746, opened 2021-10-15
update queries package and add gas limit buffers where needed

andytcf:


clementbalestrat:
@andytcf why are we passing a gasLimitBuffer everywhere? It's all very confusing to be honest

dbeal-eth:
we shouldn't need to pass gasLimiBuffer except in extremme circumstances

PR #761, opened 2021-10-28
Feature l2 debt page

0xjocke:
This PR enables the Debt page on L2. We dont have historical data on l2 so this only enables the "Global debt pool" and "Your Portfolio" section, see image.

Screen Shot 2021-10-28 at 1 06 41 pm
Screen Shot 2021-10-28 at 1 15 25 pm


0xclem (approved):
Nice one!!

PR #769, opened 2021-11-02
changes for upstream codegen changes

dbeal-eth:
due to the usage of the new codegen library on latest versions of `@synthetixio/queries`, some of the subgraph queries have
changed in name/functionality a bit.

This PR implements the necessary changes into staking to be compatible with the latest version of `@synthetixio/queries`.

in hindsight, the need to supply a `subgraphUrl` is a bit cumbersome and should be read from the context like all the other query functions, but
for the time being it isn't too much of a problem.

also, there was some duplicate functionality for `useClaimedStatus.ts` which was ultimately already provided by `useUserStakingData.ts`, so I removed.

verified functionality:
* debt page works same as before
* history page works same as before
* user claim seems to work same as before

please verify me.


pages/history.tsx
```diff
@@ -9,27 +9,79 @@ import TransactionsContainer from 'sections/history/TransactionsContainer';
 import StatsSection from 'components/StatsSection';
 
 import StatBox from 'components/StatBox';
-import useSynthetixQueries from '@synthetixio/queries';
+import { issuance, SynthetixQueryContext } from '@synthetixio/queries';
 import { useRecoilValue } from 'recoil';
 import { walletAddressState } from 'store/wallet';
+import { StakingTransactionType } from 'sections/history/types';
+import _ from 'lodash';
+import React from 'react';
```
bachstatter, line 17: Can you import `{Context}` on line one instead?

pages/history.tsx
```diff
@@ -9,27 +9,79 @@ import TransactionsContainer from 'sections/history/TransactionsContainer';
 import StatsSection from 'components/StatsSection';
 
 import StatBox from 'components/StatBox';
-import useSynthetixQueries from '@synthetixio/queries';
+import { issuance, SynthetixQueryContext } from '@synthetixio/queries';
 import { useRecoilValue } from 'recoil';
 import { walletAddressState } from 'store/wallet';
+import { StakingTransactionType } from 'sections/history/types';
+import _ from 'lodash';
```
bachstatter, line 16: To help with code splitting can you import the individual functions instead?

pages/history.tsx
```diff
@@ -9,27 +9,79 @@ import TransactionsContainer from 'sections/history/TransactionsContainer';
...
-	const feesClaimedQuery = useFeeClaimHistoryQuery(walletAddress);
 	const { setTitle } = UIContainer.useContainer();
 
-	const isLoaded = feesClaimedQuery.isSuccess;
-	const feesClaimed = feesClaimedQuery.data ?? [];
+	const isLoaded = issues.isSuccess && burns.isSuccess && feeClaims.isSuccess;
+
+	const history = isLoaded
+		? _.sortBy(
+				_.flatten([
```
bachstatter, line 63: You should be able to just use `flat` ie. `[[]].flat()`

sections/synths/components/SynthPriceCol.tsx
```diff
@@ -3,23 +3,34 @@ import styled from 'styled-components';
...
 
-import { Period } from 'constants/period';
-
 import CurrencyPrice from 'components/Currency/CurrencyPrice';
 import { CurrencyKey } from '@synthetixio/contracts-interface';
 import useSelectedPriceCurrency from 'hooks/useSelectedPriceCurrency';
 import { NO_VALUE } from 'constants/placeholder';
-import useSynthetixQueries from '@synthetixio/queries';
+import useSynthetixQueries, { exchanges, SynthetixQueryContext } from '@synthetixio/queries';
+import React from 'react';
```
bachstatter, line 11: Another one where you can just import `{useContext}`

sections/synths/components/SynthPriceCol.tsx
```diff
@@ -3,23 +3,34 @@ import styled from 'styled-components';
...
-	const historicalRates = useHistoricalRatesQuery(currencyKey, Period.ONE_DAY);
+
+	const exchangesURL =
+		React.useContext(SynthetixQueryContext)?.context.subgraphEndpoints.exchanges || '';
+	const historicalRates = exchanges.useGetRateUpdates(
+		exchangesURL,
+		{
+			orderBy: 'timestamp',
+			orderDirection: 'desc',
+			where: { timestamp_gt: Date.now() - 86400, currencyKey },
```
bachstatter, line 29: Rather using a magic number can you add a `PeriodInSeconds` in `constants/period` similar to how we have `PERIOD_IN_HOURS`

sections/synths/components/SynthPriceCol.tsx
```diff
@@ -32,7 +43,9 @@ const SynthPriceCol: FC<SynthPriceColProps> = ({ currencyKey }) => {
 					currencyKey={currencyKey}
 					price={price}
 					sign={selectedPriceCurrency.sign}
-					change={historicalRates.data?.change}
+					change={historicalRates.data?.[0].rate
+						.div(historicalRates.data?.[historicalRates.data?.length - 1].rate)
+						.toNumber()}
```
bachstatter, line 48: Can you create a function that takes historicalRates.data and returns change? Then we can add some tests and add some logic to avoid infinity

hooks/useGlobalHistoricalDebtData.ts
```diff
@@ -20,18 +21,27 @@ const useGlobalHistoricalDebtData = () => {
...
+	const isLoaded = dailyIssued.isSuccess && dailyBurned.isSuccess;
 
 	useEffect(() => {
 		if (isLoaded) {
-			const activity = activityQuery.data ?? [[], []];
+			const activity = [dailyIssued.data ?? [], dailyBurned.data ?? []];
 
 			// We concat both the events and order them (asc)
-			const eventBlocks = orderBy((activity[0] as any).concat(activity[1]), 'timestamp', 'asc');
+			const eventBlocks = orderBy((activity[0] as any).concat(activity[1]), 'id', 'asc');
```
bachstatter, line 44: Just confirming that this change has the indented behaviour. Is sorting on id the same as sorting on timestamp?

bachstatter (changes requested):
Left some comments. All the logic make sense to me!
If we're not very rushed to get this out I would think it make sense for the issuance to grab the URLS from context as you mentioned.

hooks/useGlobalHistoricalDebtData.ts
dbeal-eth: yes the old function pulled out the timestamp for you, but in this case we have to use the ID because the codegen doesn't do that.

pages/history.tsx
dbeal-eth: https://github.com/Synthetixio/staking/pull/769/commits/0a9be4a938eb698efd5a4b16244d182fdacbbbb5

pages/history.tsx
dbeal-eth: https://github.com/Synthetixio/staking/pull/769/commits/0a9be4a938eb698efd5a4b16244d182fdacbbbb5

pages/history.tsx
dbeal-eth: https://github.com/Synthetixio/staking/pull/769/commits/0a9be4a938eb698efd5a4b16244d182fdacbbbb5

sections/synths/components/SynthPriceCol.tsx
dbeal-eth: https://github.com/Synthetixio/staking/pull/769/commits/0a9be4a938eb698efd5a4b16244d182fdacbbbb5

sections/synths/components/SynthPriceCol.tsx
dbeal-eth: https://github.com/Synthetixio/staking/pull/769/commits/0a9be4a938eb698efd5a4b16244d182fdacbbbb5

sections/synths/components/SynthPriceCol.tsx
dbeal-eth: https://github.com/Synthetixio/staking/pull/769/commits/a51b4cf2e4fb959f0bac4305dfe0f93adedf9e02

dbeal-eth:
@0xclem defer to you on decision to update the queries library with better context for subgraph Url requirement in functions is worth it.

bachstatter (approved):
Looks good to me. Would be good to get @0xclem to have a look too. 
Also I didn't check it out and tested it, only reviewed based on the code.

0xcdb (approved):

0xclem (approved):

PR #771, opened 2021-11-04
Add snapshot external link to the proposals

andytcf:


bachstatter (approved):
👍
I think we need to get this merged https://github.com/Synthetixio/staking/pull/769
Or downgrade synthetix/queries before we can deploy dev. 
Pretty sure that's why the dev build is failing

PR #778, opened 2021-11-09
Fix/coinbase wallet

0xjocke:
This make sure users can connect with coinbase wallet, both when they have a browser wallet installed and when they do not. I made small and descriptive commits.

Bonus fix: I also fixed an error that we kept seeing realted to ens/address `Error: invalid ENS name`. This was due to the fact that we tried to get loans data before an address had been set/resolved.

Currently if they have metamask/browerwallet with the wrong network selected they will not be able connect.
This PR tries to tackle that #780

### DIFF
--- a/containers/Connector/Connector.tsx
+++ b/containers/Connector/Connector.tsx
@@ -61,9 +61,16 @@ const useConnector = () => {
 
 	useEffect(() => {
 		const init = async () => {
-			// TODO: need to verify we support the network
 			const networkId = await getDefaultNetworkId();
 
+			if (!window.ethereum) {
+				setAppReady(true);
+				setNetwork({ name: NetworkName.Mainnet, id: networkId, useOvm: false });
+				setSynthetixjs(synthetix({ networkId, useOvm: false }));
+				return;
+			}
+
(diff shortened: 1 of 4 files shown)


containers/Connector/Connector.tsx
```diff
@@ -61,9 +61,16 @@ const useConnector = () => {
...
+				setAppReady(true);
+				setNetwork({ name: NetworkName.Mainnet, id: networkId, useOvm: false });
+				setSynthetixjs(synthetix({ networkId, useOvm: false }));
+				return;
+			}
+
+			// TODO: need to verify we support the network
 			const provider = loadProvider({
 				networkId,
 				//infuraId: process.env.NEXT_PUBLIC_INFURA_PROJECT_ID,
```
fritzschoff, line 76: can be removed, no?

bachstatter: Will be removed in this PR: https://github.com/Synthetixio/staking/pull/780
`loadProvider` throws if given a network synthetix doesn't support

0xclem (approved):
Nice one!

PR #779, opened 2021-11-09
Make sure gas price doesn't have decimals

bachstatter:
I'm not sure if this is Kovan related but in some gases the gas price has decimals. This is causing Big number problems.
The error I saw while trying to claim was:
`Error: underflow (fault="underflow", operation="BigNumber.from", value=2000000007.0000002, code=NUMERIC_FAULT`

Just making sure that the normalised gas price doesn't have decimals seems to do the trick.
I vaguely remember @dbeal-eth  running in to this too but at the time we couldn't reproduce it.

### DIFF
--- a/sections/earn/ClaimTab/ClaimTab.tsx
+++ b/sections/earn/ClaimTab/ClaimTab.tsx
@@ -227,13 +227,15 @@ const ClaimTab: React.FC<ClaimTabProps> = ({ tradingRewards, stakingRewards, tot
 						gasLimit = normalizeGasLimit(gasLimit);
 					}
 
+					const normalizedGasPriceValue = Math.round(normalizedGasPrice(gasPrice.toNumber()));
+
 					const transaction: ethers.ContractTransaction = delegateWallet
 						? await FeePool.claimOnBehalf(delegateWallet.address, {
-								gasPrice: normalizedGasPrice(gasPrice.toNumber()),
+								gasPrice: normalizedGasPriceValue,
 								gasLimit,
 						  })
 						: await FeePool.claimFees({
-								gasPrice: normalizedGasPrice(gasPrice.toNumber()),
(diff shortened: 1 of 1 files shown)


sections/earn/ClaimTab/ClaimTab.tsx
```diff
@@ -227,13 +227,15 @@ const ClaimTab: React.FC<ClaimTabProps> = ({ tradingRewards, stakingRewards, tot
 						gasLimit = normalizeGasLimit(gasLimit);
 					}
 
+					const normalizedGasPriceValue = Math.round(normalizedGasPrice(gasPrice.toNumber()));
```
0xclem, line 230: Hmmm this is weird. I don't think this is where it should be fixed. We might have to go where the gas price is being fetched and check what's happening

0xclem (changes requested):

sections/earn/ClaimTab/ClaimTab.tsx
bachstatter: Hmm yea I agree.. Should've dug a bit deeper. We use the result of calling: `await FeePool.estimateGas.claimFees()` as gas estimate, I guess this must have returned a number with decimals.. Dont have any account I can claim with right now :(

sections/earn/ClaimTab/ClaimTab.tsx
```diff
@@ -252,6 +254,7 @@ const ClaimTab: React.FC<ClaimTabProps> = ({ tradingRewards, stakingRewards, tot
 						setTxModalOpen(false);
 					}
 				} catch (e) {
+					console.log(e);
```
fritzschoff, line 257: you need that?

bachstatter:
This got added as hot fix yesterday, so closing this

PR #780, opened 2021-11-09
Fix/handle wrong network

0xjocke:
This PR tried to handle unsupported networks better. 

If a user visit staking with an unsupported network in metamask we will ask them to swtich the ethereum mainnet.
We need this even before they are connected since we instanciate onboard before we connect.

If they are connected and switch to an unsupported network same thing.

I also added correct types for window.ethereum by installing `@metamask/providers` and removing `@metamask/detect-provider` since it was diluting the global ethereum with an incorrect type. I added this package in our `utils` folder

I'm still using the commit hash version from synthetix monorepo. If we get this https://github.com/Synthetixio/js-monorepo/pull/118 merged today I think we should do a patch release. And then I will upgrade the version here.


0xclem (approved):
Very nice but please let's make sure this PR is fully tested (as in manual testing). We need to do transactions on L1 and L2 from Metamask and hardware wallet to make sure everything is ok.

bachstatter:
Yep cool I'll make sure to test 
- Metamask
- Coinbase
- Trezor
- ledger 
On both l1 and l2.. Do you think that will cover it?

bachstatter:
I think it's all working. Basically all the same issues as mentioned here https://github.com/Synthetixio/staking/pull/781#issuecomment-964676916

Overall it's less error and correctly swiches network for you

PR #781, opened 2021-11-10
Release

0xjocke:
Release notes
- Upgrade synthetix packages, now using subgraphs
- Refactor loans page, better code quality and avoid infinity loop
- Enable Debt page on layer 2
- External link to proposals on gov page
- Fix coinbase wallet


bachstatter:
Testing done:
**Claimed on optimism mainet with:**
- Metamask + Ledger
- Metamask + Trezor

**Successfully connected to mainnet with:** 
- Coinbase
- Wallet Connect
- Ledger
- Trezor
- Metamask

No transactions made on mainnet 💸

**Claimed, mint and burn on kovan with:**
- Metmask

**Known/Found issues:**
- The subgraphs does not work on Optimism Kovan. So if you try to connect with that network we are crashing.
- If you have metamask installed and an unsupported network selected there seems to be some cause some issues when connecting with Coinbase & Wallet Connect. 
- When minting on kovan the UI is not displaying amount correctly, but the transaction goes through with out a problem. This is also happening on production.

@0xclem  Maybe you can do some testing too and then if it all looks good I'll merge this tomorrow

components/BannerManager/BannerManager.tsx
```diff
@@ -14,18 +14,23 @@ import { snapshotEndpoint } from 'constants/snapshot';
...
 	} = useSynthetixQueries();
 
 	const walletAddress = useRecoilValue(walletAddressState);
 
 	const liquidationData = useGetLiquidationDataQuery(walletAddress);
 	const debtData = useGetDebtDataQuery(walletAddress);
 	const hasVotedForElectionsQuery = useHasVotedForElectionsQuery(snapshotEndpoint, walletAddress);
-	const feeClaimHistory = useFeeClaimHistoryQuery(walletAddress);
+
+	const feeClaims = issuance.useGetFeesClaimeds(
```
fritzschoff, line 29: I'm not a native speaker, but what is claimeds?

bachstatter: I thing the code generating these hooks takes the name and adds `s` for the hook that return an array.
So the name for this is getFeesClaimed, then `use` since its a hook and `s` since it's return an array

bachstatter:
Found one more issue on the `/synths` page, happing both when connected to kovan and mainnet. 
This is from the `useGetRateUpdates` hook. 
https://github.com/Synthetixio/staking/blob/dev/sections/synths/components/SynthPriceCol.tsx#L27
I'm pretty sure we have a logic error here. `Date.now()` returns a timestamp in ms and we're subtracting one day in seconds.
I tried changing it to both ms and s. Both seems to throw an error. 
Do we have any documentation for the subgraphs? Or could you help out @dbeal-eth?

The error is:
`Error: Failed to decode `Bytes` value: `Odd number of digits``

PR #782, opened 2021-11-10
deposit and withdrawal moved to optimism gateway

dbeal-eth:
there is no need to maintain our own deposit/withdrawal page becuase optimism provides a maintained bridge for us.

this removes all deposit/withdrawal code, and instead links the appropriate points in our UI to their bridge.


JChiaramonte7 (commented):
Makes sense, but I'll leave approvals to anyone more across staking atm

constants/links.ts
```diff
@@ -29,5 +29,9 @@ export const EXTERNAL_LINKS = {
 	},
 	L2: {
 		Optimism: 'https://gateway.optimism.io/welcome',
+		SynthetixDeposit:
+			'https://gateway.optimism.io/?l1Token=0xc011a73ee8576fb46f5e1c5751ca3b9fe0af2a6f&l2Token=0x8700daec35af8ff88c16bdf0418774cb3d7599b4',
```
0xclem, line 33: We can probably use snxJS to build the URL, but definitely fine for now

0xclem (approved):
Good stuff

0xclem:
> Makes sense, lgtm, but I'll leave approvals to anyone more across staking atm

True but I can't say no to less code in staking :)

PR #785, opened 2021-11-12
Fix/use get rate updates

0xjocke:
This fixes a few issues with `useGetRateUpdates`.

Unfortunately for some reason it's in a inifinty loop, and the percentage change looks a bit off.
Didn't have time to resolve that today.. If someone wants to try to figure it out that would be great..

Edit: this has now been fixed, see https://github.com/Synthetixio/staking/pull/785#issuecomment-968570478


bachstatter:
Infinity problem fixed.  Feel free to have a look at the recent commits if curious. TLDR in a few places we gave invalidated `useMemo` when we wanted it to be memoized. We have to be a bit careful with passing `wei` as a hook dependency

0xclem (approved):

PR #795, opened 2021-11-19
isBelowCRatio not calculated correctly in userStakingData

0xjocke:
- `targetCRatio` and `targetThreshold` is both `debtData.targetCRatio`, so removed `targetThreshold`
- Added test for `calculateIsBelowCRatio` and implemented what I think is the correct calc


0xclem (approved):
Great job!

PR #796, opened 2021-11-19
Use synthetix query/mutation hooks on the claim tab

0xjocke:
This PR refactors the ClaimTab to use our Synthetix hooks. 
I think it's simpler now, less code  and we will get the changes coming from js-monorepo rather than having to remember that this piece of code was doing custom contract calls. It's based on #795 which ensures that we can rely on `isBelowCRatio`

The "Close Fee Period" button that's available on l2 when `now() > feePeriodDuration + startTime` was not implemented perfectly. The GasSelector component showing the price was always using the claim gas limit to display estimated price.
I opted to never display both those button at the same time, ie. if  the "Close Fee Period" button is visible I do not render the claim button. Do you think this make sense? I'm pretty comfortable with the code so happy to change if needed.

I've tested the claim button a fair bit. 
- Actually claiming
Made sure it handles:
- Has claimed
- No rewards
- Low C-Ratio

I did not test the "Close Fee Period" button, not sure how I would do that?

Merge plan:
Once approve we should still wait for current version of staging to be released. Then merge this and get some more testing in on staging.

### DIFF
--- a/sections/earn/ClaimTab/ClaimOrCloseFeeButton.tsx
+++ b/sections/earn/ClaimTab/ClaimOrCloseFeeButton.tsx
@@ -1,0 +1,78 @@
+import { useRouter } from 'next/router';
+import { useTranslation } from 'react-i18next';
+import { StyledButton } from '../common';
+import { Tooltip } from 'styles/common';
+import styled from 'styled-components';
+import Wei from '@synthetixio/wei';
+import ROUTES from 'constants/routes';
+
+const ClaimOrCloseFeeButton: React.FC<{
+	handleClaim: () => void;
+	handleCloseFeePeriod: () => void;
+	canClaim: boolean;
+	isBelowCRatio: boolean;
+	hasVoted: boolean;
+	hasClaimed: boolean;
+	totalRewards: Wei;
+	isCloseFeePeriodEnabled: boolean;
+}> = ({
+	handleClaim,
+	handleCloseFeePeriod,
+	canClaim,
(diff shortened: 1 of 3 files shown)


sections/earn/ClaimTab/ClaimOrCloseFeeButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+	return (
+		<Tooltip
+			hideOnClick={true}
+			arrow={true}
+			placement="bottom"
+			content={t('earn.actions.claim.ratio-notice')}
+			disabled={!canClaim || !isBelowCRatio}
+		>
+			<PaddedButtonContainer>
+				{hasVoted ? (
```
0xclem, line 46: So if `hasVoted` is true, we show the `not voted` button?

0xclem: Also can you please check with @andytcf if this needs to be disabled for L2?

0xclem (approved):
Very nice sir. Just a couple of question reg voting

sections/earn/ClaimTab/ClaimOrCloseFeeButton.tsx
bachstatter: > So if hasVoted is true, we show the not voted button?
Nice catch, that looks like a mistake
> Also can you please check with @andytcf if this needs to be disabled for L2?
Will chat to him

PR #804, opened 2021-11-24
dev => master

0xclem:
Release notes:
- If user have unsupported network selected -> switch to mainnet
- Dependency upgrades
- Amounts now being displayed when doing custom mint
- Confirmation modal incorrectly showing up when doing custom mint and then a mint max straight after fixed.
- Avoid leaving dust when max burning, both on burn tab and merge-accounts tab
- Refactor Claim tab to use synthetix-queries package
- When clicking dismiss after claiming the confirmation ui now disappears


bachstatter (approved):
Done some testing everything seems to be working fine

PR #805, opened 2021-11-25
Add dHedge swap on Uniswap to Debt-Manage tab

fritzschoff:
This PR implements an Iframe for uniswap to hedge your position with sUSD => debt mirror token


sections/debt/components/ManageTab/index.tsx
```diff
@@ -1,12 +1,34 @@
...
+import { walletAddressState } from 'store/wallet';
 import styled from 'styled-components';
-
 import { FlexDivColCentered } from 'styles/common';
 
 const ManageTab = () => {
+	const walletAddress = useRecoilValue(walletAddressState);
+	const { debtBalance: actualDebt } = useUserStakingData(walletAddress);
+
+	const uniswapURLWithQueryParams = (debt: Wei) =>
```
bachstatter, line 12: No biggie, but some ideas for small improvements.
- Name this function `getDehedgeUniswapUrl` so give some better context.
- store the address in a const, something like `dHedgeTokenAddress`
- Can also move this function out of the render method

sections/debt/components/DebtHedgingInfoPanel/DebtHedgingChart.tsx
```diff
@@ -45,7 +45,9 @@ const DebtHedgingChart: React.FC = () => {
 						interval="preserveEnd"
 						tick={{ fontSize: 10, fill: colors.white, fontFamily: fonts.mono }}
 						tickLine={false}
-						tickFormatter={(tick) => format(new Date(tick * 1000), 'd MMM yy').toUpperCase()}
+						tickFormatter={(tick) =>
+							format(new Date(tick * 1000 || new Date().getTime()), 'd MMM yy').toUpperCase()
+						}
```
bachstatter, line 50: Did we have cases where the tick produced an invalid Date? Can we apply a filter on the data to remove those data points rather than using an incorrect timestamp?

sections/debt/components/ManageTab/index.tsx
```diff
@@ -1,12 +1,34 @@
...
+		`https://app.uniswap.org/#/swap?inputCurrency=0x57ab1ec28d129707052df4df418d58a2d46d5f51&exactAmount=${debt.toNumber()}&exactField=input`;
 	return (
 		<ManageContainer>
-			Debt mirror
-			<ManageSubtitle>Coming soon</ManageSubtitle>
+			{!walletAddress ? (
+				<span>Please connect a wallet</span>
+			) : (
+				<iframe
```
bachstatter, line 19: Not sure why it's important to add a title attribute to the iframe, but lets do it to avoid this warning

sections/debt/components/DebtHedgingInfoPanel/DebtHedgingChart.tsx
fritzschoff: if you look at prod. The x axis starts at 1970.1.1. I console logged the tick and it is undefined. Dunno what causes it. That is why I went with that solution

fritzschoff: But if you think a filter function is helpful here, I can write one

sections/debt/components/ManageTab/index.tsx
fritzschoff: The token is currently unknown, waiting for the dhedge team to port the token from polygon. 
For the other two points: yes, will change

sections/debt/components/DebtHedgingInfoPanel/DebtHedgingChart.tsx
bachstatter: Hmm yea I guess we dont want to filter out data without a timestamp. 
Took a quick look, data coming from here:
https://github.com/Synthetixio/staking/blob/dev/hooks/useGlobalHistoricalDebtData.ts#L49
We could just not push if timestamp is missing

sections/debt/components/ManageTab/index.tsx
bachstatter: Yep cool!

sections/debt/components/DebtHedgingInfoPanel/DebtHedgingChart.tsx
fritzschoff: if you look at line 24 and 28 in that file, you see we don't query for timestamps. Cause the endpoint don't provide timestamps. I removed the prop entirely.
I'm also thinking, we should show maybe some graph from the uniswap pool, when it exists? Would make more sense? But not sure for now.

bachstatter:
Maybe we should host a fork of uni where sUSD isn't blacklisted

fritzschoff:
Seems a bit too much for just one token swap. Maybe we could go with just an sUSD input and we send the tx direct to the uniswap v3 router.

bachstatter:
Hey, was gonna leave a few comments but it was easier to just change the code a bit than writing a comment.. 
You can have a look at the two most recent commit and hopefully you think it's an improvement too.
- use findIndex instead of for loop
- better data strucutre for event blocks so we dont need to @ts-ignore

The chart looks great, nice work 💪

bachstatter:
> Seems a bit too much for just one token swap. Maybe we could go with just an sUSD input and we send the tx direct to the uniswap v3 router.

Yea make sense. I guess I was thinking that it could be useful to point worried users to it, but probably not worth it

0xjocke (approved):

fritzschoff:
> Hey, was gonna leave a few comments but it was easier to just change the code a bit than writing a comment.. You can have a look at the two most recent commit and hopefully you think it's an improvement too.
> 
> * use findIndex instead of for loop
> * better data strucutre for event blocks so we dont need to @ts-ignore
> 
> The chart looks great, nice work 💪

Nice, Thank you!

hooks/useGlobalHistoricalDebtData.ts
```diff
@@ -29,30 +86,76 @@ const useGlobalHistoricalDebtData = () => {
...
-				const multiplier = event.type === StakingTransactionType.Burned ? -1 : 1;
-				const aggregation =
-					data.length === 0 ? event.totalDebt : multiplier * event.value + data[i - 1].issuance;
-
-				data.push({
-					issuance: aggregation,
-					debtPool: event.totalDebt,
-					timestamp: event.timestamp,
-				});
+			let lastKnownDebtPoolPrice = new Wei(0);
```
0xclem, line 135: you can do `wei(0)`

sections/debt/components/DebtHedgingInfoPanel/DebtHedgingInfoPanel.tsx
```diff
@@ -14,7 +14,16 @@ const DebtHedgingInfoPanel: React.FC<DebtHedgingInfoPanelProps> = ({ hidden }) =
...
 		<InfoPanelContainer>
 			<InfoPanelTitle>{t('debt.actions.manage.info-panel.title')}</InfoPanelTitle>
-			<InfoPanelBody>{t('debt.actions.manage.info-panel.body')}</InfoPanelBody>
+			<InfoPanelBody>
+				{t('debt.actions.manage.info-panel.body')}
+				<br />
+				<Trans
+					i18nKey="debt.actions.manage.info-panel.link"
+					components={[
+						<ExternalLink href="https://app.dhedge.org/pool/0x65bb99e80a863e0e27ee6d09c794ed8c0be47186" />,
```
0xclem, line 23: Same here, let's put this into a `constant` file

0xclem (approved):
Great stuff @fritzschoff !

fritzschoff:
waiting for confirmation of text from @Rafanator and then we are good to go

fritzschoff:
Green light from Rafa

PR #808, opened 2021-12-02
Feat EIP1559/Optimism support

0xjocke:
- Updated to use the synthetix queries from this PR https://github.com/Synthetixio/js-monorepo/pull/122
- Calculate transaction price differently depending on mainnet, optimism and if EIP1559
- Render estimated fees correct depending on mainnet, optimism and if EIP1559
- Refactored MintTab to use the new version of GasEstimator and the GasPrice structure.

Whats left to do:
Refactor more components to use the new version of GasEstimator and the GasPrice structure.
Agree on how we want the UI to look.
Currently looks like this:
Screen Shot 2021-12-02 at 4 39 18 pm
Screen Shot 2021-12-02 at 4 39 29 pm


utils/network.ts
```diff
@@ -18,14 +19,32 @@ export async function getDefaultNetworkId(): Promise<NetworkId> {
...
+
 export const getTransactionPrice = (
-	gasPrice: Wei | null,
+	gasPrice: GasPrice | null,
 	gasLimit: GasLimitEstimate,
-	ethPrice: Wei | null
+	ethPrice: Wei | null,
+	optimismLayerOneFee: Wei | null
 ) => {
 	if (!gasPrice || !gasLimit || !ethPrice) return null;
```
0xclem, line 37: maybe we can return 0? Not sure

utils/network.ts
```diff
@@ -18,14 +19,32 @@ export async function getDefaultNetworkId(): Promise<NetworkId> {
...
+	const totalGasPrice = getGweiGasPriceOrMaxFee(gasPrice);
+
+	const extraLayer1Fees = optimismLayerOneFee?.mul(GWEI_UNIT); // optimismL1Fees comes in ETH
+	const txPrice = totalGasPrice
```
0xclem, line 41: Really good!

0xclem (approved):

utils/network.ts
bachstatter: hmm yea,  I wonder how it renders. As you can see I didn't touch that line.  Ideally it should show some indication that we're not ready to show the price yet.

bachstatter:
I've now converted all places where useSynthetixTx was used.
I haven't done too much testing, notes here:
**BurnTab** 
not executed but estimates correctly
**ClaimTab**
 not executed but estimates correctly

**Transfer modal:**
Something is wrong when doing a transfer. Do you have any insights why the gaslimit would be `unpredictable`?  If you don’t I’ll try and figure it out.
```
require manual gas limit (error={"code":-32000,"message":"execution reverted"}, method="estimateGas", transaction={"from":"0xfe3068C4c65Ec3Cb53b3864Aa19c19D21Df13840","maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x17b400956a"},"maxFeePerGas":{"type":"BigNumber","hex":"0x30566c52d4"},"to":"0xfE18be6b3Bd88A2D2A7f928d00292E7a9963CfC6","data":"0xb014c3a30000000000000000000000006e9f5370a9d9bd9d4bb667ef9aeae665b8e9b17f000000000000000000000000000000000000000000000000000000231d264508","accessList":null}, code=UNPREDICTABLE_GAS_LIMIT, version=providers/5.4.5)
```

**RedeemableDeprecatedSynthsModal**
Need some guidance of how I can test this.

Merge account nominate
not executed but estimates correctly

**MergeActionBox.tsx Txn: 'RewardEscrowV2', 'mergeAccount'**
Not tested, I think I need to nominate account first?

**Merge Account burn**
Estimated correct but not executed


**Layer Migrate tab**
Not tested

**Staking rewards**
Estimated correct but not executed

All the components under the **LP Tab** is not using `useSynthetixTxn`. So since that requires some more refactoring I'll make that work in another PR, based on this branch.

bachstatter:
@0xclem If you have any thoughts about the `Transfer modal` problems I'm all ears

0xclem:
> @0xclem If you have any thoughts about the `Transfer modal` problems I'm all ears

No idea. What are you trying to transfer? Maybe you don't any transferrable token?

0xclem (approved):
@bachstatter this is really good! Well done

utils/network.ts
bachstatter:
```
const transactionFee = getTransactionPrice(
  gasPrice,
  gasLimitEstimate,
  ethPriceRate,
  optimismLayerOneFee
);
...
transactionFee != null
  ? `(${formatCurrency(selectedPriceCurrency.name, transactionFee, {
	  sign: selectedPriceCurrency.sign,
	  })})`
  : ''
```
At the call site we are handling `null` by not showing anything, which I guess is slightly better than showing 0 incorrectly

PR #825, opened 2021-12-15
fix governance page

andytcf:


bachstatter (approved):
I am a little worried about the package upgrades if we haven't done much testing. 
But I guess we can do some testing on staging?

PR #833, opened 2021-12-16
Deploy governance changes to master

andytcf:
/gov was outdated since snapshot started updating stuff - for the election i had to make changes in the jsmonorepo for governance and add the new council mapping so people can vote on staking


bachstatter (approved):
I did some quick testing and made sure voting works. Seems to be working fine!

PR #850, opened 2021-12-20
fix for: indexing_error

fritzschoff:
Bildschirmfoto 2021-12-20 um 15 18 45


bachstatter (approved):
I just made sure that we keep the v2 of package-lockfile. Ran `npm i` with npm version 7

PR #852, opened 2021-12-22
fix price calculation for snx graph

fritzschoff:
the app used an old and not working query.
For the stats page we might need pagination? maybe `first:100` isn't enough


bachstatter (approved):

bachstatter:
Tooltip:
image

If my changes make sense, feel free too merge.

PR #864, opened 2022-01-13
L2 loans

bachstatter:
This PR enables loans UI for L2. The loans code was in a fairly bad state with missing/wrong types so it's also a fair amount of refactor.

The code is extra complex cause there are different contracts on L2 vs L1. I've tried to put much of that logic in functions rather than in the component/ giant useEffect. I might write some tests tomorrow, but it should be ready for review.

It's also unfortunate that the subgraphs doesn't work for loans on L2, so to get the open loans I had to parse the event logs.  They should be ready in a few weeks, at that point we could simplify the `getOpenLoans` function. 

Sorry for enormous PR. I've split things up into smallish commits. If something doesn't make sense fire away questions and then I can put some more comments in the code too.


containers/Loans/Loans.tsx
```diff
@@ -56,133 +56,67 @@ function Container() {
...
-
-				const fetchLoan = async (owner: string, id: string) => {
-					return makeLoan({
-						loan: await loanStateContract.getLoan(owner, id),
-						type,
-						minCRatio: await loanContract.minCratio(),
+			loanContracts.forEach((loanContract, i) => {
+				if (!loanContract) return;
+				const loanStateContract = loanStateContracts[i];
+				const updateLoan = async (owner: string, id: BigNumber) => {
```
fritzschoff, line 112: maybe a `Promise.all` is better here

containers/Loans/Loans.tsx
```diff
@@ -195,36 +129,34 @@ function Container() {
...
 					);
 					await updateLoan(owner, id);
 				};
 
-				const onCollateralWithdrawn = async (
-					owner: string,
-					id: string,
-					amount: ethers.BigNumber
-				) => {
+				const onCollateralWithdrawn = async (owner: string, id: BigNumber, amount: BigNumber) => {
```
fritzschoff, line 159: why async func?

containers/Loans/Loans.tsx
bachstatter: That big `forEach` is just setting up events and calling appropriate functions when the events happening. No async stuff happening

containers/Loans/Loans.tsx
bachstatter: All of the event handlers that updates the loan are async functions. You can see it `await updateLoan(owner, id);` a few lines below.. It is a fair point that it isn't really necessary since we don't care about the result of calling updateLoan. But it also doesn't hurt and at least communicates that the function is doing something asynchronous.

Happy to remove async + await for those functions if you feel strongly about it, have other reasoning I'm not thinking about

bachstatter:
@fritzschoff  If you have time it would good to get an approval tick on this one too

containers/Loans/Loans.tsx
fritzschoff: no was just wondering. all good thanks :)

containers/Loans/getOpenLoans.tsx
```diff
@@ -0,0 +1,90 @@
+import { BigNumber, ethers, providers } from 'ethers';
+import { Network } from 'store/wallet';
+import { getInfuraRpcURL } from 'utils/infura';
+import getLoan from './getLoan';
+import { Loan } from './types';
+
+/**
+ * Note, once the subgraphs has synced up on optimism it would make sense to use those instead the events log
```
fritzschoff, line 8: isn't the graph synced on L2?

fritzschoff (approved):

containers/Loans/getOpenLoans.tsx
bachstatter: Yea there were some issue and noah said it would take a few weeks to sycn

fritzschoff: yes but I heard that only affects the graph on L1 but that could be already old information

bachstatter: I think I discovered the issue, this query should return closed and open loans:
```
{
  loans(first: 5) {
    id
    isOpen
  }
}
```
And I get an empty array back https://thegraph.com/hosted-service/subgraph/synthetixio-team/optimism-main?selected=playground

PR #865, opened 2022-01-15
Update staking dApp with new queries package release and gov-ovm support

andytcf:
This PR concludes https://sips.synthetix.io/sips/sip-191/ which deployed a new contract on ovm, this PR connects the staking dApp and allows us to consume the new contract using ovm providers. Other features include:

1. No need to login with wallet to view proposals anymore
2. Pagination on votes more than 1000 count

---

### Flows to test

- [ ] (L1 & L2) When a user is not connected to the staking dApp - they should still be able to view all proposals + results and voting history across all FOUR spaces
- [ ] (L1 & L2) When a user visits a proposal with more than 1000 votes they should be ABLE to see more than 1000 votes in the count
- [ ] (L1) When a user visits the /gov/snxgov.eth page with their wallet logged in - they should NOT be able to create a proposal - it should say switch to L2 or something
- [ ] (L2) When a user visits the /gov/snxgov.eth page with their wallet logged in - they should be ABLE to create a proposal


sections/gov/hooks/useCouncilMembers.ts
```diff
@@ -15,16 +15,20 @@ export const useCouncilMembers = () => {
 	useEffect(() => {
 		const getCouncilMembers = async () => {
 			if (provider && isAppReady) {
+				const L1Provider = new ethers.providers.InfuraProvider(1, process.env.INFURA_ARCHIVE_KEY);
+
 				let spartanCouncilContract = new ethers.Contract(
 					spartanCouncil.address,
 					spartanCouncil.abi,
-					provider as ethers.providers.Provider
+					L1Provider as ethers.providers.Provider
```
bachstatter, line 23: No need to use `as` here. It is already the correct type

bachstatter (approved):
Left some minor comments.. The `isWalletConnected` is the only one that might be important

PR #880, opened 2022-01-24
Wallet connect optimism

bachstatter:
Ensure we have rpc urls defined for both mainnet and optimism


containers/Connector/config.ts
```diff
@@ -56,7 +56,18 @@ export const initOnboard = (
...
 					walletName: 'walletConnect',
-					rpc: { [network ? network.id : networkId]: infuraRpc },
+					rpc: {
+						[NetworkId['Mainnet']]: getInfuraRpcURL({
+							name: Network['Mainnet'],
+							id: NetworkId['Mainnet'],
+							useOvm: false,
+						}),
+						[NetworkId['Mainnet-Ovm']]: getInfuraRpcURL({
+							name: Network['Mainnet'],
```
bachstatter, line 66: Using `mainnet` rather than `Mainnet-Ovm` on purpose. The `Network` enum has `Mainnet-Ovm` defined as `mainnet-ovm` infura wants `mainnet-optimsim`.  `getInfureRpcUrl` will append optimism when `useOvm` is set to true

bachstatter:
Vercel should be happy once your guni PR is merged

containers/Connector/config.ts
fritzschoff: shouldn't we update the `getINfuraRpcUrl` func?

fritzschoff (approved):

containers/Connector/config.ts
fritzschoff: I think that is how the new NetworkId is going to look like `export const NetworkId = {
	Mainnet: 1,
	Ropsten: 3,
	Rinkeby: 4,
	Goerli: 5,
	Kovan: 42,
	'Mainnet-Ovm': 10,
	'Kovan-Ovm': 69,
	'Local-Ovm': 420,
	Local: 31337,
} as const;

export type NetworkIdType = typeof NetworkId;`

PR #891, opened 2022-01-27
Implementation of handleSwitchChain function

fritzschoff:
Due to the new releases of the js-monorepo I also had to update the types.
So this PR isn't all about the handleSwitchChain function implementation. 
⚠️ I also removed the prefix from the `getInfuraURL` function that caused an issue here: https://github.com/Synthetixio/staking/pull/880
close https://github.com/Synthetixio/issues/issues/403


bachstatter (approved):

bachstatter:
I've added some more commits, feel free to have a look.  @fritzschoff

PR #900, opened 2022-02-01
Fix/more ren btc fixes

bachstatter:
I finally got my hands on some renBTC and could sort out all issues.
We have our our own fake renBTC deployed on kovan


fritzschoff (approved):
LGTM

PR #901, opened 2022-02-01
Release

bachstatter:
- ENS name support
- renBTC loans fixed
- APY calculation fixed
- dDhege debt hedging UI live
- Watch wallet issue fixed


package.json
```diff
@@ -25,7 +26,7 @@
 		"@synthetixio/contracts-interface": "2.57.3",
 		"@synthetixio/optimism-networks": "2.57.3",
 		"@synthetixio/providers": "2.57.3",
-		"@synthetixio/queries": "2.57.3",
+		"@synthetixio/queries": "0.0.0-b64ea26",
```
bachstatter, line 29: Ideally we should release js-monorepo instead, but we're awaiting a review. See https://github.com/Synthetixio/js-monorepo/pull/159/files

PR #905, opened 2022-02-03
useLockedSnxQuery instead of useGlobalStakingInfoQuery

bachstatter:
useGlobalStakingInfoQuery was unnecessarily complicated (and have some kind of bug).
We only used `stakedValue`. So using the new `useLockedSnxQuery` to get that value instead

Note: at the time of writing no one has called `mint` so the snx rewards on mainnet is 0 leading to an APY of 0.5% (sUSD fees only)


package.json
```diff
@@ -26,7 +26,7 @@
 		"@synthetixio/contracts-interface": "2.57.3",
 		"@synthetixio/optimism-networks": "2.57.3",
 		"@synthetixio/providers": "2.57.3",
-		"@synthetixio/queries": "0.0.0-b64ea26",
+		"@synthetixio/queries": "0.0.0-8fc31c6",
```
fritzschoff, line 29: Can yo switch to the latest version if you use different query?

fritzschoff (approved):
Well done! But I don't get why we had something on L2 but not sure n L1

PR #909, opened 2022-02-04
Ens avatar support

sasicodes:
- Shows ens avatar only if the user is already set. For users with no avatar, it won't be displayed.

No libraries installed.

image

### DIFF
--- a/sections/shared/Layout/UserMenu/UserMenu.tsx
+++ b/sections/shared/Layout/UserMenu/UserMenu.tsx
@@ -231,12 +231,12 @@ const Ens = styled.span`
 	display: flex;
 	align-items: center;
 	text-transform: lowercase;
-	gap: 5px;
 `;
 
 const EnsAvatar = styled.img`
 	height: 15px;
 	border-radius: 100px;
+	margin-right: 5px;
 `;
 
 const NetworkTag = styled(FlexDivCentered)`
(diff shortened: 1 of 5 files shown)


sasicodes:
@fritzschoff Can you please review this 🙇🏻

sections/shared/Layout/UserMenu/UserMenu.tsx
```diff
@@ -219,8 +227,16 @@ const Menu = styled.div`
 	grid-auto-flow: column;
 `;
 
-const EnsName = styled.span`
+const Ens = styled.span`
+	display: flex;
+	align-items: center;
 	text-transform: lowercase;
+	gap: 5px;
```
fritzschoff, line 234: unfortunately gap is not supported in safari when using flexbox. 
There are not many users that connect via safari to our dapp but still, would be annoying for them. 
Can you make the spacing with margin? Or something else please ?

fritzschoff (changes requested):

fritzschoff:
@sasicodes thank very much! Small minor issue. After that we are good to go :)

sections/shared/Layout/UserMenu/UserMenu.tsx
sasicodes: sure

sasicodes:
@fritzschoff Updated ser!

fritzschoff:
Awesome. Thanks so much!

PR #913, opened 2022-02-07
Fixed unused prop for pie chart comp

fritzschoff:


bachstatter (approved):
Added a commit to avoid type casting (using `as`).
Other than that, nice!

PR #1006, opened 2022-04-01
added custom pie chart data

fritzschoff:
@bachstatter created the synhts by hand so that Afif and Kaleb can update them by hand with a PR. Build is failing because of some type errors in the wei lib. Dunno why. Rewrote the js monorepo in npm workspace without lerna so once this is merged we should be fine


sections/debt/components/DebtPieChart/DebtPieChart.tsx
```diff
@@ -47,10 +50,192 @@ type DebtPieChartProps = {
 };
 
 const SynthsPieChart: FC<DebtPieChartProps> = ({ data, isLoaded, isLoading }) => {
-	const totalSupply = data ? data : undefined;
+	//	const totalSupply = data ? data : undefined;
 
-	const pieData = useMemo(() => {
-		const supplyData = totalSupply?.supplyData ?? [];
+	const pieData = [
```
bachstatter, line 55: Lets put this in a seperate file so its easier to update

bachstatter (changes requested):
As I said in previous comment, lets put the json in a seperate file.
Can you also clean up some of the lint error?
I also think we should add a note somewhere saying "Data last updated 2022-01-01 22:30"  where the date is the date the JSON was created.

I'll be online tonight we we can have a catchup about these changes

PR #1016, opened 2022-04-09
[FEAT] Desktop sidebar: show tooltip info about total SNX staked

kennym:
# Description

image

I was confused about this number, and I believe more users could feel the same. Adding some clarity would help I believe.

This PR fixes that.

# Screenshot

image


bachstatter:
Thanks!

PR #1017, opened 2022-04-10
[FIX] Table component: remove code smell

kennym:
Addresses Linter warning "Ternary operators should not be nested (typescript:S3358)"


bachstatter:
Hey! Thanks for the PR. The format is a little bit wrong, could you make sure to run prettier as well? 
In general we dont have this eslint rule enabled. I think to actually make this code less smelly we should create a `SortableTableHead` component.  Then we could have an early return for `column.isSorted` and simple/single if statement for the `column.isSortedDesc` logic

kennym:
> Hey! Thanks for the PR. The format is a little bit wrong, could you make sure to run prettier as well? In general we dont have this eslint rule enabled. I think to actually make this code less smelly we should create a `SortableTableHead` component. Then we could have an early return for `column.isSorted` and simple/single if statement for the `column.isSortedDesc` logic

thanks, I'll work on that over the weekend. Ping you when it's ready.

kennym:
@bachstatter done

components/Table/SortTableHead.tsx
```diff
@@ -0,0 +1,59 @@
...
+				<StyledSortIcon src={SortUpIcon} viewBox={`0 0 ${SortUpIcon.width} ${SortUpIcon.height}`} />
+				<StyledSortIcon
+					src={SortDownIcon}
+					viewBox={`0 0 ${SortDownIcon.width} ${SortDownIcon.height}`}
+				/>
+			</SortIconContainer>
+		);
+	}
+
+	if (isSorted && isSortedDesc) {
```
kennym, line 29: I'm pretty sure the `isSorted` check is needed.. not sure why it's complaining.

bachstatter: Since you do an early return on line 18 when isSorted is false we know that isSorted is gonna be true here. 
```
if (!isSorted) {
		return (...)
```
So you can remove it!

kennym: true! Done.

bachstatter (approved):
Nice work!

PR #1028, opened 2022-04-19
Perf updates

bachstatter:
This PR adds three performance improvements
1. Change refetch interval to 5 min instead of every 30 sec.
 Queries will still update as you navigate around, we could even consider disable this

2. We dont need to refetch everytime the user focus the window
Making all queries refetch every time the window refocus is putting a lot of load on the app.
Instead of having this as a default for att queries it should be used specifically for queries that want this.
Most of the time I dont think it's needed since the data will update when user navigates around

3.  When calculating the 7day rate change use the end of hour price 7 days ago
The `sevenDaysAgoSeconds` gets passed to `subgraph.useGetRateUpdates`
The subgraph hooks pass all arguments as query keys. This meant we were constantly rerendering since the `now` timestamp kept changing


fritzschoff (approved):
Well done!

PR #1030, opened 2022-04-21
Fix errors in console

bachstatter:
Fixes `Warning: Expected server HTML to contain a matching <div> in <div>` and ` Error: invalid address or ENS name (argument="name", value=null...`

### DIFF
--- a/constants/NotificationContainer.tsx
+++ b/constants/NotificationContainer.tsx
@@ -2,9 +2,14 @@ import { createPortal } from 'react-dom';
 import styled from 'styled-components';
 import { ToastContainer } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.css';
+import { useState, useEffect } from 'react';
 
 const NotificationContainer = () => {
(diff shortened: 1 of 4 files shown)


fritzschoff (approved):

constants/NotificationContainer.tsx
```diff
@@ -2,9 +2,14 @@ import { createPortal } from 'react-dom';
...
 import 'react-toastify/dist/ReactToastify.css';
+import { useState, useEffect } from 'react';
 
 const NotificationContainer = () => {
-	return typeof document !== 'undefined'
+	const [mounted, setMounted] = useState(false);
+	useEffect(() => {
+		setMounted(true);
+	}, []);
+	return mounted
```
Rickk137, line 12: let's move to a seprate hook like useMounted and reuse it in the other place too

Rickk137 (changes requested):

Rickk137 (approved):

PR #1035, opened 2022-05-03
Grab open loans from subgraphs

bachstatter:


containers/Loans/Loans.tsx
```diff
@@ -71,6 +72,21 @@ function Container() {
...
+	);
+
+	const subgraphOpenLoansKey = subgraphOpenLoansQuery.data
+		? JSON.stringify(
+				subgraphOpenLoansQuery.data.map((x) => {
+					const id = x.id.replace(/-\w+/, ''); //remove -sETH from id
+					return { ...x, id };
+				})
+		  )
+		: '';
```
Rickk137, line 88: I don't understand the reason behind stringifing the array and then parsing it down in the code

containers/Loans/Loans.tsx
```diff
@@ -91,11 +107,13 @@ function Container() {
...
-				loanContracts,
-				loanStateContracts,
+				erc20LoanContract,
+				ethLoanContract,
+				erc20LoanStateContract,
+				ethLoanStateContract,
 				address,
-				network,
 				isL2,
+				subgraphOpenLoans: subgraphOpenLoansKey ? JSON.parse(subgraphOpenLoansKey) : [],
```
Rickk137, line 116: here we are parsing the subgraphOpenLoansKey. we could simply store it as array instead of stringify

Rickk137, line 116: here we are parsing the subgraphOpenLoansKey. we could simply store it as array instead of stringify

Rickk137 (changes requested):

containers/Loans/Loans.tsx
bachstatter: We want to rerun this useEffect when there is a change in openLoans. You cant put and array in the `useEffect`s dependency array

containers/Loans/Loans.tsx
bachstatter: See comment above

containers/Loans/Loans.tsx
Rickk137: Nice, didn't notice that

Rickk137 (approved):

PR #1054, opened 2022-05-16
Delegate action

bachstatter:
When changing permission or adding a new delegate wallet it doesn't show up on the table.
The reason for this is cause the table's data is fetch from a subgraph and the subgraph doesn't have time to get synced.

I try to resolve this in two ways. 
- `sleep` for 5s before quering the subgraph
-  Add refresh button with a tooltip explanation why a refresh is required
image


fritzschoff (approved):
Lgtm

PR #1087, opened 2022-05-27
Self liquidation UI

0xjocke:
This adds a new "Self Liquidation" tab to the staking page.
This tab will show up if:
1.  User is staking
2. User c-ration is below target c-ratio
3. User sUSD balance is less than burn amount to fix c-ratio
4. it's not a delegate wallet

The UI it self will look different depending on if you have any sUSD.
If your sUSD balance is bigger than 0:
Screen Shot 2022-05-27 at 2 32 02 pm

If your balance is 0:
Screen Shot 2022-05-27 at 2 31 10 pm


fritzschoff (approved):

noisekit (approved):
Looking good to me. The only suggestion that would simplify the code and probably improve UX: keep the tab available but if user does not match the criteria - show different content (explain what self-liquidation is and why you don't need it?)

So no need for any redirections and makes this new functionality discoverable.

PR #1090, opened 2022-05-27
Self liquidate icon

0xjocke:
- Changes the tab component so we can pass in `color` rather than `blue:true`.  Most tabs to use `blue:true` so I also made the default color blue.
- Add warning icon for self liquidation tab 

Screen Shot 2022-05-27 at 4 53 04 pm


nkbt (approved):
I guess the actual icon will come later after refactoring

noisekit (approved):

bachstatter:
It's there in the second commit. Using a warning icon

PR #1096, opened 2022-06-02
Self liquidation side panel

0xjocke:
Adds a side panel with calculation of how your collateral/ debt / c-ratio will change from the self liquidation ui.
There's a few different calc:
When burn max button is visible:
Screen Shot 2022-06-02 at 3 57 48 pm
When self liquidation button is visible and we have enough non escrowed SNX
Screen Shot 2022-06-02 at 1 16 24 pm
When self liquidation button is visible and we DONT have enough non escrowed SNX
Screen Shot 2022-06-02 at 1 16 46 pm
No button visible:
Screen Shot 2022-06-02 at 3 54 44 pm


noisekit (approved):
Looks good to me in the UI. The code is all good as far as I can understand it too!

PR #1099, opened 2022-06-06
Bundle report

noisekit:
- [x] Bundle analyzer for visual bundle inspection
- [x] Removed extra webpack 5 dependency (Next uses webpack 4 for build and we do not use webpack anywhere ourselves)

# Running reporting
Gives pointers on which files were generated

image

# Server report
image

# Browser report
image


0xjocke (approved):
Nice! As discussed I think yes to a lot of the upgrade questions :D

PR #1115, opened 2022-06-12
Gov/tab fixes

jmzwar:


sections/gov/components/Panel.tsx
```diff
@@ -128,9 +127,9 @@ const Panel: React.FC<PanelProps> = () => {
 							setProposal(null);
 							setPanelType(PanelType.LIST);
 							if (activeTab === SPACE_KEY.PROPOSAL) {
-								router.push(ROUTES.Gov.Home);
+								router.back();
```
bachstatter, line 130: Since both cases are calling back we wouldn't need the if statements. Does this work if I reload the page while on a proposal though? Or just visiting a proposal directly?

jmzwar: Yup - updated.

bachstatter (approved):

PR #1117, opened 2022-06-13
[NOT READY TO MERGE] Use burnToTarget in selfLiquidation burn max button

0xjocke:
- We do this to avoid the minStake time check in the Issuer contract

### DIFF
--- a/sections/staking/components/SelfLiquidateTab/BurnMaxButton.tsx
+++ b/sections/staking/components/SelfLiquidateTab/BurnMaxButton.tsx
@@ -9,18 +9,34 @@ import styled from 'styled-components';
 import { Synths } from 'constants/currency';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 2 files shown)


bachstatter:
This fix required quite a lot of understanding context of contracts and some edgecases. Probably worth having a call about this one @jmzwar

noisekit (approved):
I reckon the check if button needs to be rendered should be outside of the button itself, but as a quick fix looks good to me.

bachstatter:
> I reckon the check if button needs to be rendered should be outside of the button itself, but as a quick fix looks good to me.

Yea there is a check there. It wont render the button if that's the case. 
I just added it as an extra safeguard since the button is called Burn Max but is actually calling `burnToTarget`.
Either way we still have some things to figure out. The contract wasn't as flexible as I hoped so it might be that we have to remove this button all together :(

bachstatter:
Replaced by #1126

PR #1125, opened 2022-06-16
[MERGE AFTER ELECTION] After election updates

jmzwar:
This PR
- Removes the dilution logic from the proposal section of governance.
- Fetches the current spartan council from the new Election Module contract


bachstatter (approved):
Looks good

sections/gov/components/Proposal/DilutionContent.tsx
```diff
@@ -1,9 +1,13 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
```
andytcf, line 1: I dont think we need dilution anymore

andytcf: or rename this file, pretty sure we have one called "Content.tsx" already

andytcf (approved):
added some comments

sections/gov/components/Proposal/DilutionContent.tsx
bachstatter: Yea should rename. We actually removed Content.tsx, since it was only used for the election table that no longer exists

bachstatter:
The snapshot is now reading from the new election contract. This means that current spartan concil cant vote untill the election closes in a few hours. I also means that we can merge this one.

PR #1133, opened 2022-06-17
Browser support

noisekit:
- [x] Update browser databases
- [x] Limit browser support to the latest versions
- [x] Add CI check to ensure we keep browsers db up to date
- [x] Adjust babel config for Jest (speed up things), but also include all files in test coverage report

Locally everything seems to work, will do more testing on the preview.
UPD: Preview app smoketest went OK (minted some susd without issues)

### DIFF
--- a/components/Input/NumericInput.tsx
+++ b/components/Input/NumericInput.tsx
@@ -1,4 +1,4 @@
-import { ChangeEvent, FC } from 'react';
+import type { ChangeEvent, FC } from 'react';
 import styled from 'styled-components';
 import { FixedNumber } from 'ethers';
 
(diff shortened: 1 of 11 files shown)


components/Input/NumericInput.tsx
```diff
@@ -1,4 +1,5 @@
-import { ChangeEvent, FC } from 'react';
+import React from 'react';
+import type { ChangeEvent, FC } from 'react';
```
bachstatter, line 2: Is `import type` this required for the babel test config?

0xjocke (approved):

jmzwar (approved):

PR #1136, opened 2022-06-20
Prettify the code

noisekit:
- [x] Ignore more things in prettier and eslint configs
- [x] Apply Prettier to all the files
- [x] Add Prettier check to CI
- [x] Switch to use lint-staged (much faster than checking full code) and check with lint and prettier on commits


jmzwar (commented):
Looks ok, but let's check it doesn't clash with eslint. I thing there is already a prettier plugin.

jmzwar (approved):

0xjocke (approved):

PR #1145, opened 2022-06-26
Remove headerTitle, subTitle from context

jmzwar:
This PR removes logic for setting the Header title and subtitle state. Title/subtitle are to be replaced by derived values from the nav state.


bachstatter (approved):
Looks good, just fix some lint errors

PR #1152, opened 2022-06-29
Update build and testing dependencies

noisekit:
- Should make tests faster as we no longer include/process all the images
- removed few older deps
- updated to the latest jest, eslint, etc

NOTE: extracted from #1100 (did there in attempt to solve dependencies audit issues)


test-setup/svgMock.js
```diff
@@ -1 +1,9 @@
 module.exports = () => null;
```
bachstatter, line 1: Can remove this line?

0xjocke (approved):
One minor comment, otherwise looks good

bachstatter:
Oh just saw this was a draft..

PR #1158, opened 2022-07-03
Feature/additional clean

jmzwar:
This PR pulls out the logic for adding optimism into a hook, as well as deals with much of the re-rendering issues occurring in the navigation on both mobile and desktop by splitting out these parts into separate components.

demo: [here](https://staking-9b3jgcnf4-synthetixio.vercel.app/)


sections/shared/hooks/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+						},
+					],
+				});
+			} catch (e) {
+				setNetworkError((e as Record<'message', string>).message);
+			}
+		}
+
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [provider, network?.id]);
```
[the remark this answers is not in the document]
bachstatter: Both `setNetworkError` and  `t` are functions that doesn't change so we can add them as deps without any problems. 

Can also them remove `/* eslint-disable react-hooks/exhaustive-deps */`

sections/shared/Layout/SideNav/Mobile/MobileMenu.tsx
```diff
@@ -0,0 +1,132 @@
...
+		`}
+	}
+
+	&:after {
+		width: 2px;
+		height: 40px;
+		content: '';
+		position: absolute;
+		top: 0;
+		/* the line needs to outside (so around -3px), however due to overflow issues, it needs to be inside for now */
```
bachstatter, line 120: This comment is out of date?

bachstatter (approved):
Left some minor comments. Looks good! 
I think we might need to coordinate a bit with #1166. If that gets merge first we need to switch how we render svg.

containers/UI/reducer.ts
```diff
@@ -0,0 +1,47 @@
...
+				activeMobileSubNav: action.subMenu,
+			};
+		case 'clear_sub':
+			return {
+				...state,
+				isMobileSubNavOpen: false,
+				activeMobileSubNav: null,
+			};
+		default:
+			return { ...state };
```
bachstatter, line 45: A bit of a nit pick but no need to spread here

bachstatter (approved):

PR #1160, opened 2022-07-06
Optimise contracts by cleaning up deployments JSON

noisekit:
- Replace synthetix (resolves to precompiled `browser.js` by default) with synthetix source so we can control compilation
- Stub imports of unsupported networks (we only support mainnet and kovan + ovm)
- Cleanup deployments.json for each supported network (cleanup unused fields and replace abi with non-json readable format)

On a clean `test.js` file importing synthetix resulsted in this:
```sh
# before:
asset test.out.js 9.51 MiB [emitted] (name: main)

# after:
asset test.out.js 2.59 MiB [emitted] (name: main)
```

## Big browserjs chunk is gone! Size is down significantly

Well, it is still massive, but improvements is definitely there!

image

PS: Need to extract synthetix into its own chunk btw. Seems like there is a bit too much duplication

## Testing on Kovan
image

### DIFF
--- a/scripts/minify-synthetix-contract.js
+++ b/scripts/minify-synthetix-contract.js
@@ -1,0 +1,22 @@
+const fs = require('fs');
+const ethers = require('ethers');
+
+module.exports = function generate({ networks, out }) {
+	networks.forEach((network) => {
+		const deployment = require(`synthetix/publish/deployed/${network}/deployment.json`);
+
+		Object.keys(deployment.targets).forEach((key) => {
+			const { name, source, address } = deployment.targets[key];
+			deployment.targets[key] = { name, source, address };
+		});
+
+		Object.keys(deployment.sources).forEach((key) => {
+			const source = deployment.sources[key];
+			const iface = new ethers.utils.Interface(source.abi);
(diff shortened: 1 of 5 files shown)


jmzwar (approved):

scripts/minify-synthetix-contract.js
```diff
@@ -0,0 +1,22 @@
...
+		Object.keys(deployment.targets).forEach((key) => {
+			const { name, source, address } = deployment.targets[key];
+			deployment.targets[key] = { name, source, address };
+		});
+
+		Object.keys(deployment.sources).forEach((key) => {
+			const source = deployment.sources[key];
+			const iface = new ethers.utils.Interface(source.abi);
+			const abi = iface.format(ethers.utils.FormatTypes.full);
+			deployment.sources[key] = { abi };
```
bachstatter, line 17: Do we do this to strip out stuff that we don't need? I would expect the a oi to already be there

0xjocke (approved):
This is so good. I'm not at my computer today so only did a quick review. I could do a more in depth tomorrow or next week. But if it works I guess it doesn't hurt to get it out to staging. 

Nice work 💪

PR #1166, opened 2022-07-08
Import svg and images directly instead of using react-optimized-image wrapper

noisekit:
Preview: https://staking-git-noisekit-synthetixio.vercel.app/

- apply `svgo` to all svg images (halfed the size for many)
- Add `@svgr` webpack loader
- walk over every usage of Svg/Img and replace with Component + width
- Remove all the optimised images related plugins and libs
- Cleanup configs
- Update Next build to handle static assets

NOTE: when width/height matches viewBox, they are cleaned up and because in many places we rely on the svg size for some reason - instead of checking every single place I have used `width` just everywhere. Ideally SVG needs to be rendered inside some container or properly styled with css, so it does not try to take all the space possible and instead rendered inside the limited box.

## Some random screenshots (visually identical to prod)
Screen Shot 2022-07-11 at 12 30 14
Screen Shot 2022-07-11 at 12 30 26
Screen Shot 2022-07-11 at 12 32 52
Screen Shot 2022-07-11 at 12 32 57
Screen Shot 2022-07-11 at 12 33 04
Screen Shot 2022-07-11 at 12 33 10
Screen Shot 2022-07-11 at 12 35 48


sections/delegate/DelegateTable.tsx
```diff
@@ -111,12 +110,10 @@ const DelegateTable: FC = () => {
...
 								<NoMarginP>
-									<Svg
+									<Refresh
+										width="16"
 										style={{ cursor: 'pointer' }}
-										src={Refresh}
-										onClick={() => {
-											delegateWalletsQuery.refetch();
-										}}
+										onClick={() => delegateWalletsQuery.refetch()}
```
jmzwar, line 116: We'll need to test these just to be safe

jmzwar: That the onClick area has not changed.

0xjocke (approved):
Looked around I think it all looks good. Did one small commit fixing the width on one icon

jmzwar (approved):
Looks good.

PR #1170, opened 2022-07-11
Cleanup unused libs

noisekit:
Instead of loading 2 css + 1 font - added (cleaned up) styles for slick carousel right in place.

Looks same with much less code

image

LIVE at https://staking-git-noisekit-synthetixio.vercel.app/


jmzwar (approved):
I've tested on desktop and mobile and looks good.

PR #1174, opened 2022-07-12
Release

bachstatter:
- Refactor and speed improvements
- Dependency upgrades
- Some minor content changes


bachstatter:
Not sure if it's been tested enough. Quite a lot of changes already though so creating a relase PR so that dev doesn't grow too much

jmzwar:
I've some done basic testing and haven't run into anything huge. Nikita and I did find one issue with loading pages when chaging routes, but he has put another PR in for that.

jmzwar (approved):
Can't see any issues but perhaps would be better to leave for a couple of days so we can test claiming etc.

PR #1175, opened 2022-07-13
Reduce content popping

noisekit:
1. Disable server side rendering
2. Update body background to match app
3. Create GlobalLoader (cannot import `svg` as Next is "too smart" and prohibits importing "images" for pre-generated html, not understanding that it is actually imported as react component).
4. Add `loading` state for dynamic imports
5. `safe-import` is unfortunately removed (again because Next is "too smart" and does replace internal importing functionality preventing it throwing exceptions)

We still have some things to do about that initial styling on dashboard page

Loader freezes for me because JS processing eats up all my CPU 🤦 
loader-state

### DIFF
--- a/components/GlobalLoader/GlobalLoader.tsx
+++ b/components/GlobalLoader/GlobalLoader.tsx
@@ -1,0 +1,50 @@
+import Colors from 'styles/theme/colors';
+
+export function GlobalLoader() {
+	const size = 50;
+	return (
+		<svg
+			viewBox="0 0 38 38"
+			xmlns="http://www.w3.org/2000/svg"
+			style={{
+				width: `${size}px`,
+				height: `${size}px`,
+				left: '50%',
+				top: '50%',
+				marginTop: `-${size / 2}px`,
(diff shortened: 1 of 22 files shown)


pages/_document.tsx
```diff
@@ -44,6 +46,16 @@ export default class MyDocument extends Document {
 			<Html lang="en">
 				<Head>
 					<style type="text/css" dangerouslySetInnerHTML={{ __html: mediaStyles }} />
+					<style
+						type="text/css"
+						dangerouslySetInnerHTML={{
+							__html: `
```
jmzwar, line 52: Can we do this in the main css file?

components/GlobalLoader/GlobalLoader.tsx
```diff
@@ -0,0 +1,50 @@
+import Colors from 'styles/theme/colors';
+
+export function GlobalLoader() {
+	const size = 50;
```
jmzwar, line 4: Nice

jmzwar (approved):

PR #1189, opened 2022-07-25
minor bug fixes

jmzwar:
Disable the transactions feature and fix the account switching feature.


0xjocke (approved):
I thought I tested the account switching already. But I might only have tested in metamask

PR #1195, opened 2022-07-27
Remove erc20 loans

0xjocke:


jmzwar (approved):
Can't see any issues here, but not across the full context with loans. Will do some testing now.

PR #1204, opened 2022-08-01
Pull liquidation info from the new subgraph

noisekit:
POC
- works on optimism-mainnet
- works on optimism-kovan
- TODO: periodical checks on timer
- TODO: support liquidation subgraphs on mainnet

image

NOTE: I don't think we should put all the subgraphs into a single repo, maybe monorepo? So we can treat each of them as a separate package.
Right now I just ran all the queries directly from staking, but we could collocate those queries with the subgraph code itself as an option...

### DIFF
--- a/components/BannerManager/Liquidation/useLiquidation.ts
+++ b/components/BannerManager/Liquidation/useLiquidation.ts
@@ -1,0 +1,34 @@
+import { wei } from '@synthetixio/wei';
+import useSynthetixQueries from '@synthetixio/queries';
+import Connector from 'containers/Connector';
+import { useLiquidationOptimismSubgraph } from './useLiquidationOptimismSubgraph';
+
+export function useLiquidation() {
+	const { useGetLiquidationDataQuery, useGetDebtDataQuery } = useSynthetixQueries();
+	const { walletAddress } = Connector.useContainer();
+
+	const liquidationData = useGetLiquidationDataQuery(walletAddress);
+	const debtData = useGetDebtDataQuery(walletAddress);
+
+	const issuanceRatio = debtData?.data?.targetCRatio ?? wei(0);
+	const cRatio = debtData?.data?.currentCRatio ?? wei(0);
+	const liquidationDeadlineForAccount =
+		liquidationData?.data?.liquidationDeadlineForAccount ?? wei(0);
+
+	const ratio = issuanceRatio.eq(0) ? 0 : 100 / Number(issuanceRatio);
+
+	const optimismDeadline = useLiquidationOptimismSubgraph();
+	const deadline =
+		optimismDeadline > 0
+			? optimismDeadline
+			: Number(liquidationDeadlineForAccount.toString()) * 1000;
+
+	const hasWarning =
+		optimismDeadline > 0 || (!liquidationDeadlineForAccount.eq(0) && cRatio.gt(issuanceRatio));
+
+	return {
+		hasWarning,
+		ratio,
+		deadline,
+	};
+}
--- a/components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
+++ b/components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
@@ -8,7 +8,7 @@ function getEndpoint(networkName: String) {
 		case 'mainnet-ovm':
 			return 'https://api.thegraph.com/subgraphs/name/noisekit/liquidator-optimism';
 		default:
-			return null;
+			throw Error(`Called with unsupported network: ${networkName}`);
 	}
 }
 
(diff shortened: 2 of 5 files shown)


components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
```diff
@@ -0,0 +1,81 @@
...
+		}
+	);
+}
+
+export function useLiquidationOptimismSubgraph(queryOptions?: UseQueryOptions) {
+	const { data, isSuccess } = useLiquidationOptimismInfo(queryOptions);
+	if (!isSuccess) {
+		return 0;
+	}
+	// @ts-ignore I give up
```
fritzschoff, line 79: :D what?

components/BannerManager/Liquidation/useLiquidation.ts
```diff
@@ -0,0 +1,34 @@
...
+	const liquidationDeadlineForAccount =
+		liquidationData?.data?.liquidationDeadlineForAccount ?? wei(0);
+
+	const ratio = issuanceRatio.eq(0) ? 0 : 100 / Number(issuanceRatio);
+
+	const optimismDeadline = useLiquidationOptimismSubgraph();
+	const deadline =
+		optimismDeadline > 0
+			? optimismDeadline
+			: Number(liquidationDeadlineForAccount.toString()) * 1000;
```
fritzschoff, line 24: if `liquidationDeadlineForAccount` is `wei`. I guess you can directly do `liq.toNumber() * 1000`

components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
```diff
@@ -0,0 +1,81 @@
+import { useQuery, UseQueryOptions } from 'react-query';
+import Connector from 'containers/Connector';
+
+function getEndpoint(networkName: String) {
+	switch (networkName) {
+		case 'kovan-ovm':
+			return 'https://api.thegraph.com/subgraphs/name/noisekit/liquidator-optimism-kovan';
```
bachstatter, line 7: It would be good to add this subgraph to https://github.com/synthetixio/synthetix-subgraph/ instead. 

We should probably also remove the old liquidation graph: https://github.com/Synthetixio/synthetix-subgraph/blob/main/subgraphs/liquidations.js

If it's added to that repo we can then pull/refresh the subgraph [main.json](https://github.com/Synthetixio/js-monorepo/blob/master/packages/queries/subgraphs/main.json) file in the queries lib, and it will generate typed hooks for us . 

It's a bit more work. But also good for you to get to know that repo. And convenient when upgrades to the liquidation contract happens, good to have one place to remember to update

bachstatter: Also when we do a deploy from that repo re deployes the subgraphs to all supported network. And the hooks will use the correct url automatically

components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
bachstatter: Made a PR showing how you can type it:
https://github.com/Synthetixio/staking/pull/1205

bachstatter:
> NOTE: I don't think we should put all the subgraphs into a single repo, maybe monorepo? So we can treat each of them as a separate package.
Right now I just ran all the queries directly from staking, but we could collocate those queries with the subgraph code itself as an option...

I missed this note before adding my other comments. I do the benefits I described above is worth having it in synthetix-subgraph.. But happy to have a chat about it

components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
bachstatter: > The graph generates a different structure now and fitting it with the old way does not make much sense to me
You mean the liquidation graph? I think the current liquidation graph in unused.

bachstatter: > > And the hooks will use the correct network url in the hooks automatically from context.
> 
> That's the part I would like to avoid. We already depend too much on abstractions that cannot be untied right now, trying to run all queries without relying on any of the existing libs and context that includes all the contracts.
> 
> That's kind of the point of this POC

Yea I guess to would be nice to decouple it a little. But having hooks generated make sense to me, since with graphql you can ask for what ever parameters you need either way. 

Also, I guess decoupling it from the queries lib isn't really related to having the subgraph definition live in synthetix-subgraphs? That repo is essentially a monorepo for all our subgraphs already? And it has some smart things in the deployment script. Using the "Grafting" feature, deploying to all supported synthetix networks, deploying the mainnet graph to the decentralised network. Not saying it's perfect put there's quite a few useful features there.

bachstatter: Yea know it's a draft. Spark some good convos either way. And I do agree with your points.. I guess what's annoying is that we have some quite tricky bugs related to synth supply and SNXHolder that are more important than the liquidation subgraph. So the main benefit of you starting to look into the synthetix-subgraphs repo would be to help debugging those.. 

With that said, maybe the easiest way to resolve those bugs (in [handleTransferSynth](https://github.com/Synthetixio/synthetix-subgraph/blob/main/src/fragments/balances.ts) is to start with a fresh subgraph...

https://github.com/Synthetixio/synthetix-subgraph/blob/main/src/fragments/balances.ts

bachstatter:
> TODO: periodical checks on timer

A final note, react query have support for this :)
```
		{
			enabled: Boolean(walletAddress && network?.name),
			refetchInterval: 3000,
			...queryOptions,
		}
```

PR #1205, opened 2022-08-01
Ts example of hook

bachstatter:
Not to be merge, just sharing some knowledge of how to type react-query hooks

### DIFF
--- a/components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
+++ b/components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
@@ -8,7 +8,7 @@ function getEndpoint(networkName: String) {
 		case 'mainnet-ovm':
 			return 'https://api.thegraph.com/subgraphs/name/noisekit/liquidator-optimism';
 		default:
-			return null;
+			throw Error(`Called with unsupported network: ${networkName}`);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
```diff
@@ -52,30 +46,41 @@ export async function fetchLiquidationInfo(
...
 	});
 
-	const { data, errors } = await body.json();
+	const {
+		errors,
+		data,
+	}: {
+		errors?: Error[];
+		data: { stakerEntity: StakerEntity };
+	} = await body.json();
```
bachstatter, line 55: > ok, that's where I was missing another StakerEntity... wow

Yea and making sure this fetch fn have a proper type. By default this will be `any`. The proper way of doing it would be to rely on a runtime validation library like zod. But this works alright if the API is stable

bachstatter: Yea. The generic is needed if you want consumers of the hook to pass in options. If you have the " inner callback fetch function" typed and have the options configured directly in your hook you dont need the generic at all.

components/BannerManager/Liquidation/useLiquidationOptimismSubgraph.ts
```diff
@@ -52,30 +46,41 @@ export async function fetchLiquidationInfo(
...
+
 	if (errors?.[0]) {
 		throw new Error(errors?.[0]?.message || 'Unknown server error');
 	}
 	return data?.stakerEntity;
 }
 
-export function useLiquidationOptimismInfo(queryOptions?: UseQueryOptions) {
+export function useLiquidationOptimismInfo(queryOptions?: UseQueryOptions<StakerEntity>) {
```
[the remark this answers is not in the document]
bachstatter: Yea its a bit weird. As I wrote in that other comment it is good to infer if you dont let consumer pass in the option. But if you pass in options that you have typed it will use those options to "create" the return type. (I think)

PR #1226, opened 2022-08-11
Connecting to unsupported network

0xjocke:
This PR adds support to be connected to an unsupported network. `synthetixjs` cannot be instantiated with an unsupported network. So we can now have a signer and still have `synthetixjs` be null.
This gives us two benefits:
1. The bridge page works with swapping from unsupported networks
2. We can render a UI letting the user know that they are on the wrong network.

Most of the logic is in the first commits. Then I just went through all pages and switched out `<Connect Wallet>` button to `<ConnectOrSwitchNetwork />`.

Sorry for large PR....
The header button now say "Unsupported network"  and render buttons for switching:
Screen Shot 2022-08-11 at 2 14 08 pm

The bridge page works even when on an unsupported network:
Screen Shot 2022-08-11 at 2 13 48 pm

All other pages looks like this:

Screen Shot 2022-08-11 at 3 49 27 pm
Screen Shot 2022-08-11 at 3 48 57 pm

I tried to test all different stage before committing and I have a bunch of more screenshots if you're interested hehe...

### DIFF
--- a/constants/queryKeys.ts
+++ b/constants/queryKeys.ts
@@ -5,13 +5,13 @@ import { Synths } from './currency';
 
 export const QUERY_KEYS = {
 	Debt: {
-		WalletDebtData: (walletAddress: string, networkId: NetworkId) => [
+		WalletDebtData: (walletAddress: string, networkId: number) => [
 			'debt',
 			'walletDebtData',
 			walletAddress,
 			networkId,
 		],
-		DebtSnapshot: (walletAddress: string, networkId: NetworkId) => [
+		DebtSnapshot: (walletAddress: string, networkId: number) => [
 			'debt',
 			'debtSnapshot',
 			walletAddress,
@@ -19,7 +19,7 @@ export const QUERY_KEYS = {
 		],
 	},
 	Liquidations: {
-		LiquidationsData: (walletAddress: string, networkId: NetworkId) => [
+		LiquidationsData: (walletAddress: string, networkId: number) => [
 			'liquidations',
 			'liquidationsData',
 			walletAddress,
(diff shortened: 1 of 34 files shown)


noisekit (approved):

constants/queryKeys.ts
```diff
@@ -1,25 +1,24 @@
-import { NetworkId } from '@synthetixio/contracts-interface';
 import { WeiSource } from '@synthetixio/wei';
 import { SPACE_KEY } from 'constants/snapshot';
 import { Synths } from './currency';
 
 export const QUERY_KEYS = {
 	Debt: {
-		WalletDebtData: (walletAddress: string, networkId: NetworkId) => [
+		WalletDebtData: (walletAddress: string, networkId: number) => [
```
jmzwar, line 7: Do we need to initialise the query context if the network is not defined?

jmzwar (approved):
Nice. I've tested network switching locally and no issues on that front. Haven't done a full test of all queries/mutations however.

constants/queryKeys.ts
bachstatter: hmm not totally sure I understand the question. This file (which I dont think I've ever used before) is for defining cache keys for `useQuery`, quite smart to have that in one place. Then you can invalidate and refetch specific queries without having access to `refetch`. So instead of passing around a `refetch` function we could `queryClient.invalidateQueries(QUERY_KEYS.Debt.WalletDebtData(address, networkId))`

jmzwar: Yes I think you are right for this file specifically. Good to know (similar to apollo).

content/App.tsx
```diff
@@ -32,23 +33,23 @@ const queryClient = new QueryClient({
...
-					provider && network?.id
+					provider && isSupportedNetworkId(networkId) && synthetixjs
 						? createQueryContext({
 								provider: provider,
 								signer: signer || undefined,
-								networkId: network.id,
+								networkId,
 						  })
 						: createQueryContext({
 								networkId: 1,
```
jmzwar, line 55: This will run when network id is not supported - do we want that to occur? Shouldn't no queries run if we are not on a supported network

bachstatter: Not the clearest code. But since the tenary is this:`provider && isSupportedNetworkId(networkId) && synthetixjs` 
It means we will be on a supported network since otherwise synthetixjs would be `undefined`

PR #1230, opened 2022-08-15
Feature/speed upgrades

jmzwar:
This PR reduces the amount of contracts we are importing by only selecting those used in queries or in staking itself. The same context structure is used in Connector.tsx but the initialisation is performed by a function defined locally, not pulling from the synthetix from contracts-interface.


noisekit (approved):
Looking good, as discussed we can follow up with:
1. moving contracts to codegen in monorepo, so no need in much copypasta
2. async import contracts

content/PoolsPage.tsx
```diff
@@ -24,15 +23,15 @@ function Pool() {
   const [stakedTokens, setStakedTokens] = useState(BigNumber.from(0));
   const { t } = useTranslation();
 
-  const { walletAddress } = Connector.useContainer();
+  const { walletAddress, synthetixjs } = Connector.useContainer();
 
-  const snx = synthetix({ networkId: NetworkIdByName['mainnet-ovm'], useOvm: true });
```
bachstatter, line 29: I'm not sure this page is active. But here we're are changing behaviour. Before syntehtix was always initialised with optimism as network. Where as now it will be initialised with whatever network you are connected too

containers/Connector/Connector.tsx
```diff
@@ -71,52 +74,57 @@ const useConnector = () => {
...
-    }
-  }, []);
+        const provider = new ethers.providers.Web3Provider(update.wallets[0].provider, {
+          name: network.name,
+          chainId: networkId,
+        });
+
+        const signer = provider.getSigner();
+        const contracts = setSynthetix(networkId, signer);
+        const synthetixjs = isSupported ? ({ contracts } as SynthetixJS) : null;
```
bachstatter, line 103: Here we set `synthetixjs` to just be the `{contracts}`. I played around with the types a little (removed `any`) and noticed that just a few lines before we try to create a  `synthsMap` and  `tokensMap` by accessing `synthetixjs.synths` and `synthetixjs.tokens`. These would now be undefined, right?

content/App.tsx
```diff
@@ -50,10 +50,12 @@ const InnerApp: FC<AppProps> = ({ Component, pageProps }) => {
                 provider: provider,
                 signer: signer || undefined,
                 networkId,
+                synthetixjs,
```
bachstatter, line 53: Also, since synthetixjs now just in `{contracts: ContractsMap}` we need to make sure that every hook + function in `@synthetixio/queries` only access `ctx.snxjs.contracts`. Did a quick search and found `ctx.snxjs.toBytes32` which would fail.

I think if we wanna do this change we should type the `createQueryContext` to accept `{contracts:`SynthetixJS['contracts']` instead

containers/Connector/Connector.tsx
jmzwar: Yep good pickup

content/PoolsPage.tsx
jmzwar: This contract is only on optimism mainnet, so in useGetUniswapStakingRewardsAPY we throw an error, but maybe for clarity's sake should just hardcode for optimism here.

containers/Connector/Connector.tsx
```diff
@@ -71,52 +74,57 @@ const useConnector = () => {
...
+          },
+        });
+
+        const connectedWallets = update.wallets.map(({ label }) => label);
+        localStorage.setItem(LOCAL_STORAGE_KEYS.SELECTED_WALLET, JSON.stringify(connectedWallets));
+      } else {
+        dispatch({ type: AppEvents.WALLET_DISCONNECTED });
+      }
+    },
+    [setSynthetix]
```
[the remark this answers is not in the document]
bachstatter: Removed `setSynthetix` completely

containers/Connector/Connector.tsx
bachstatter: Merged fixes to resolve this

content/App.tsx
bachstatter: ☑️

content/PoolsPage.tsx
bachstatter: Yea I think it's fine since it also currently isn't used.

PR #1232, opened 2022-08-15
Correctly set the address when adding non ens wallet

0xjocke:


noisekit (approved):
the logic with mutable resolved address is not easy to understand, but as a fix looks like should work ok

jmzwar (approved):

PR #1235, opened 2022-08-17
Align lint with monorepo

noisekit:
Copy config from monorepo + minor adjustments for v2/ui specifically (to avoid too many changes)
Autofix + prettier


jmzwar (approved):
Approved - in future it might almost be worth moving this into it's own package

PR #1236, opened 2022-08-17
Cleanup dependencies

noisekit:
As part of monorepo migration I have discovered a lot of unused (and also missing) dependencies.

To simplify further import making this a separate PR

- [x] add depcheck with config (note that depcheck intentionally does not support `.basePath` TS aliases so we need to manually ignore them)
- [x] Remove e2e test runner and CI config (these are broken for a long time and should be re-configured after monorepo migration)
- [x] Extra: enabled recursive deps audit in CI (was temporarily disabled due to deep nested deps issues in the past)

❗❗❗ NOTE: deployed to https://staking-git-noisekit-synthetixio.vercel.app/ 🔗 

I did test burning, dsnx->susd sell, unstaking. No particular issues noticed (optimism on trust wallet)


package.json
```diff
@@ -79,21 +72,20 @@
     "remarkable": "^2.0.1",
     "styled-components": "^5.3.5",
     "styled-media-query": "^2.1.2",
+    "synthetix": "^2.0.0",
```
bachstatter, line 75: hmm are we using synthetix directly anywhere?

bachstatter: ah I see.  Does this fail if we were to remove synthetix?

package.json
```diff
@@ -110,29 +102,25 @@
...
     "prettier": "^2.7.1",
-    "react-test-renderer": "^17.0.2",
     "remarkable-external-link": "^1.1.0",
-    "typescript": "^4.7.4",
-    "url-loader": "^4.1.1"
+    "typescript": "^4.7.4"
   },
   "resolutions": {
-    "eslint-plugin-prettier": "^4.0.0",
+    "synthetix": "^2.0.0",
```
bachstatter, line 120: Why did you need to add a resolution for synthetix?

0xjocke (approved):

PR #1237, opened 2022-08-18
Speed upgrades typed

bachstatter:
This PR is based on `feature/speed-upgrades` and adds correct type for the new Synthetixjs
I think we want it to look something like this. This still have some type errors we need to resolve. 

1. Fix `synthsMap` and `tokenMap`, dont think we rely too much on those so maybe we can just refactor away usage?
2. Type https://github.com/Synthetixio/js-monorepo/blob/master/packages/queries/src/index.ts#L70 to be `synthetixjs: {contracts: SynthetixJS['contracts']} | null;`


containers/Connector/Connector.tsx
```diff
@@ -48,88 +59,52 @@ const useConnector = () => {
...
-    [L1DefaultProvider]
-  );
+      const provider = new ethers.providers.Web3Provider(update.wallets[0].provider, {
+        name: network.name,
+        chainId: networkId,
+      });
+
+      const signer = provider.getSigner();
+      const contracts = isSupportedNetworkId(networkId)
+        ? initializeSynthetix(networkId, signer)
```
jmzwar, line 83: Minor gripe but this will be re-referenced every render

bachstatter: Hmm yea but it is in a useCallback? So only when the state of onboard changes. 
I think that's only when account is changes -> new signer or when network changes. In both these cases we want "synthetix"/ the contracts to be re-initialised?

jmzwar: Ah yep - my bad - updateState is already wrapper :)

jmzwar (approved):

PR #1245, opened 2022-08-22
Alert component

fritzschoff:


components/InputBadge/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+      </Badge>
+    </StyledAnimatedWrapper>
+  );
+}
+
+const StyledAnimatedWrapper = styled(motion.div)`
+  min-width: 170px;
+  border: 1px solid black;
+  cursor: pointer;
+`;
```
bachstatter, line 38: Same comments as on the last PR, lets figure out how to style this without styled components

fritzschoff: removed it in the last PR. Will be resolved here as well when merged. This branch is branched off the other one.

components/Alert/index.tsx
```diff
@@ -0,0 +1,29 @@
+import { InfoIcon } from '@chakra-ui/icons';
+import { Container, Fade, Text } from '@chakra-ui/react';
+
+interface AlterProps {
```
jmzwar, line 4: Typo

fritzschoff: good catch, thanks

components/Alert/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  text: string;
+}
+
+export default function Alert({ text }: AlterProps) {
```
jmzwar: Good start on this. Looking at this design [here](https://www.figma.com/file/uFHKfLFYczHnZ7voSluRvO/Chakra-UI-SNX-System?node-id=517%3A9808), it looks like we have primary, secondary, tertiary variants and success, info, warning and error states.

jmzwar: Also haven't had a convo with @bachstatter or @noisekit about it, but it would be good by default if we have the option to pass a style object to all our primitives

fritzschoff: Ah didn't knew that this file existed! I was doing the component based of the burn flow in a different file. Well there is plenty to add here. Will do, thanks!

PR #1268, opened 2022-08-30
Load analytics script

noisekit:
Looks like Next does not update Head with Scripts when it is dynamically loaded and not pre-compiled on the initial load

image

### DIFF
--- a/content/App.tsx
+++ b/content/App.tsx
@@ -21,7 +21,6 @@ import SystemStatus from 'sections/shared/SystemStatus';
 
 import '../i18n';
 import Connector from 'containers/Connector';
-import Script from 'next/script';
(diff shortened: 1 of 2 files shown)


jmzwar (approved):

content/App.tsx
```diff
@@ -116,26 +115,6 @@ const App: FC<AppProps> = (props) => {
         <meta name="twitter:image" content="/images/staking-twitter.jpg" />
         <meta name="twitter:url" content="https://staking.synthetix.io" />
         <link rel="icon" href="/images/favicon.ico" />
```
bachstatter, line 117: Should all the tags in `<Head/>` be moved to `pages/_document.tsx`?

PR #1278, opened 2022-09-02
Main action cards

0xjocke:
I was gonna add some cypress component tests, but I think it's better to wait until we've migrated to js-monorepo. And maybe even until we removed next, since the configuration seems a bit different when using next vs raw webpack

https://staking-storybook-git-main-action-cards-synthetixio.vercel.app/?path=/story/mainactioncard--primary
image


v2-components/main-action-card/MainActionCard.tsx
```diff
@@ -0,0 +1,235 @@
+import { Badge, Box, Button, Flex, Heading, Stack, Text } from '@chakra-ui/react';
+import React, { PropsWithChildren, ReactNode } from 'react';
+import { getVariant } from '../c-ratio-health-card/getVariant';
```
bachstatter, line 3: This function is probably gonna be important in a few places. Maybe we should think about a better name and put it elsewhere.

bachstatter:
hmm a little more work on the disabled states coming

noisekit (approved):

v2-components/main-action-card/MainActionCard.tsx
jmzwar: Yeah for sure. Maybe something like solvency/health state?

v2-components/main-action-card/MainActionCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+        icon={<MaintainIcon />}
+      />
+      {isStaking && (
+        <Badge
+          color={variant}
+          bg="blackAlpha.600"
+          border={'1px'}
+          borderColor={variant}
+          display="flex"
+          alignItems={'center'}
```
jmzwar, line 109: It's a bit annoying but can we do just "center" and "1px"
My editor is code completing to {'center'} as well but bit annoying tbh

v2-components/main-action-card/MainActionCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+          width="fit-content"
+          margin={'0 auto'}
+          fontSize="x-small"
+          borderRadius={'base'}
+          fontWeight="700"
+        >
+          <InfoIcon color={variant} width="10px" height="10px" />
+          <Text ml="0.5">
+            {variant !== 'success'
+              ? 'Adjust to collect weekly rewards'
```
jmzwar, line 119: Should this be in i18 json?

v2-components/main-action-card/MainActionCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+    targetCratioPercentage,
+    currentCRatioPercentage,
+  });
+  const canClaim = !hasClaimed && variant === 'success';
+  return (
+    <Container>
+      <CardHeader
+        step={3}
+        headingText="Collect Rewards"
+        bodyText="Maintain your target C-Ratio to collect your weekly rewards."
```
jmzwar, line 166: +i18

jmzwar (changes requested):
Some little things

v2-components/main-action-card/MainActionCard.tsx
bachstatter: yea I dont care too much. But if we wanna enforce it I'll add `'react/jsx-curly-brace-presence': ['warn', { props: 'never', children: 'never' }],`. Then eslint can fix all places for us. 78 places in total.. Probably wait with fixing that until migraiton to monorepo is done

v2-components/main-action-card/MainActionCard.tsx
bachstatter: You caught me. My other components are also missing i18.

v2-components/main-action-card/MainActionCard.tsx
bachstatter: Just keep it in v2-components for now?  Maybe `getHealthVariant` since we have color and components variants defined with these values?

